PR #1, opened 2022-02-24
FS 536 assessment

harryyo:


app/assess/data.py
```diff
@@ -0,0 +1,107 @@
...
+        for fund in response:
+            funds.append(Fund.from_json(fund))
+        return funds
+    return None
+
+
+def get_fund(fund_id: str) -> Fund | None:
+    endpoint = FUND_STORE_API_HOST + FUND_ENDPOINT.format(fund_id=fund_id)
+    response = get_data(endpoint)
+    if response and "name" in response:
```
SentientHousePlant, line 61: At risk of being exiled by the python community (sorry folx) I actually think this would be a good use case for the match case syntax introduced in python 3.10 (which we are using!)

Here is a tutorial (a straight up pep) https://www.python.org/dev/peps/pep-0636/

app/assess/models/round.py
```diff
@@ -0,0 +1,47 @@
+from datetime import datetime
+from typing import List
+
+from slugify import slugify
+
+from .application import Application
+
+
+class Round(object):
```
SentientHousePlant, line 9: Nice data models! 

You can deduce boilerplate by using data classes (in the standard library) or attrs (a data class library which is used by basically 90% of python people, me included). 

see this [video](https://www.youtube.com/watch?v=vBH6GRJ1REM)

build.py
```diff
@@ -0,0 +1,60 @@
+import os
+import shutil
+import urllib.request
+import zipfile
+
+
+def build_assets():
```
SentientHousePlant, line 7: Great code here for building assets.

SentientHousePlant (changes requested):

app/assess/data.py
```diff
@@ -0,0 +1,107 @@
...
+        FLASK_ROOT, "tests", "api_data", "endpoint_data.json"
+    )
+    fp = open(api_data_json)
+    api_data = json.load(fp)
+    fp.close()
+    if endpoint in api_data:
+        return api_data.get(endpoint)
+
+
+def get_funds() -> List[Fund] | None:
```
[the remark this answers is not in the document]
harryyo: Ah thanks!

build.py
harryyo: Nice, agreed

app/assess/models/round.py
harryyo: Great video, on it

app/assess/data.py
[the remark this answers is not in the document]
harryyo: Thanks!

app/assess/assess.py
```diff
@@ -0,0 +1,113 @@
+from app.assess.data import get_application
+from app.assess.data import get_fund
+from flask import abort
+from flask import redirect
+from flask import render_template
+from flask import url_for
+from flask.views import MethodView
+
+
+class AssessQuestionView(MethodView):
```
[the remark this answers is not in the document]
harryyo: Totally must be core maintainer - no other explanation.

app/assess/data.py
harryyo: Great recommendation would love to use these - thanks for the intro. Though have a question. Here we want to create a fund (regardless of whether there are any rounds), and just add rounds if there are any present, so are nested if statements actually ok here? i.e. the fund has to exist to create a round so if we matched  cases on has_rounds=true/false, we'd need two cases and to create fund in both(duplicating that statement)? Or is there a way to use (or pass) an object created in another case?

app/assess/models/round.py
harryyo: Dataclasses done

SentientHousePlant (approved):
Very nice, and modern python! Great use of data classes to clean up code and reduce code duplication. 💯

app/create_app.py
```diff
@@ -0,0 +1,97 @@
...
+
+    from app.default.routes import default_bp, not_found, internal_server_error
+    from app.assess.routes import assess_bp
+    from app.assess.assess import AssessQuestionView
+
+    flask_app.register_error_handler(404, not_found)
+    flask_app.register_error_handler(500, internal_server_error)
+    flask_app.register_blueprint(default_bp)
+    flask_app.register_blueprint(assess_bp)
+    flask_app.add_url_rule(
```
Adam-W1, line 78: Very nifty approach to handling more complex route logic. 
*Just a note based on our conversation earlier - we could do with putting some thought as a team to decide an approach for how we move forward and grow this application using Flasks Url Rules (to avoid pollution of the top level py file).

Adam-W1 (approved):
Some great Pythonic code here - Bravo!

stevewalton28 (approved):

PR #3, opened 2022-03-11
FS-640 add rounds

harryyo:


app/config.py
```diff
@@ -19,11 +19,15 @@
 APIs Config
```
SentientHousePlant, line 19: The get function isn't a method of environ but rather a method used on any dictionary. Environ is merely a dictionary of environment variables. This leads the following exciting possibility...default values. 

The second argument in a get function is the default value: so you don't have to use "or" anymore.

So:

`environ.get("FUND_STORE_API_HOST") or TEST_FUND_STORE_API_HOST`

becomes 

`environ.get("FUND_STORE_API_HOST", TEST_FUND_STORE_API_HOST)`

SentientHousePlant (changes requested):

tests/test_initialisation.py
```diff
@@ -13,8 +13,8 @@ def test_flask_initiates(flask_test_client):
     GIVEN Our Flask Application
     WHEN the '/' page (index) is requested (GET)
     THEN check that the get response is successful.
-    If this test succeedes then our flask application
-    is ATLEAST up and running without errors.
+    If this test succeeds then our flask application
+    is AT LEAST up and running without errors.
     """
```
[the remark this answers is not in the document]
harryyo: Thanks.. 🍶

app/assess/models/round.py
```diff
@@ -0,0 +1,39 @@
+from dataclasses import dataclass
+from datetime import datetime
+from typing import List
+
```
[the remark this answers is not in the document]
harryyo: Great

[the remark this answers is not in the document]
harryyo: Thanks

app/config.py
harryyo: Ah-ha...! Yes get() does provide a default option... But the current expression, I propose is (marginally) preferable as a slightly more strict default - as it returns the default also when "FUND_STORE_API_HOST" is set, but to an unusable empty string, or other falsy value.

Although this will undoubtably create arguments/wars in the community on whether an empty string should be considered false.. obv!

DEMO
`environ = {"test": ""}`
`print(environ.get("test") or "nice fallback")`
`print(environ.get("test","might not get the fallback"))`

Adam-W1: got to go with the OR operand solution (i agree we want to use the default for falsy's) with Jacks nice bracket formatting suggestion to improve readability :)

Adam-W1: for env vars where we can guarantee the fallback/default value is suitable *

SentientHousePlant (approved):
It like it

README.md
```diff
@@ -64,16 +64,17 @@ You should see the following:
 
 # Configuration
 
-This service is designed to be an interface to two primary data stores: The *Fund Store* which holds data about different funds, and the *Application Store* which holds individual applications/submissions for each fund.
+This service is designed to be an interface to three primary data stores: The *Fund Store* which holds data about different funds, the *Round Store* which holds information about the rounds for each fund, and the *Application Store* which holds individual applications/submissions for each fund.
 
-The locations of the APIs for these two stores can be set using the FUND_STORE_API_ROOT and APPLICATION_STORE_API_ROOT environment variables. Set these to the respective locations for each API:
+The locations of the APIs for these two stores can be set using the FUND_STORE_API_HOST,  ROUND_STORE_API_HOST and APPLICATION_STORE_API_HOST environment variables. Set these to the respective locations for each API:
```
Adam-W1, line 69: i think should be three*

app/assess/models/fund.py
```diff
@@ -1,16 +1,15 @@
...
 
 @dataclass
 class Fund:
     name: str
     _identifier: str = None
-    applications: List[Application] = None
+    rounds: List[Round] = None
 
     @property
     def identifier(self):
```
Adam-W1, line 15: do we still need @property still?

README.md
harryyo: Great spot! Thanks.

app/assess/models/fund.py
harryyo: I agree, you're right, every fund should have an id already here. Removing.

PR #4, opened 2022-03-20
FS-689 task page

SentientHousePlant:
# Task page with minimal data added.

This PR adds a task page with all the necessary data passed into it. 

One can see that fund_data, round_id, and assessment_data are all passed into the jinja template, this will allows you to do whatever you want with the data - including producing a table and dynamic "you've completed x of y" questions line.

If javascript is enabled then a small script tag (with nonce attribute so that it is safe) will remove the "open" attribute from the details on the page. Meaning that if javascript is disabled then the details element will be open.

I added a single content check - it checks that the only piece of dynamic content on the page (the application id) is correctly passed into the project reference.

Nothing controversial in this PR that I think needs discussing - although one may want to query my choice of url. 

- [x] Harry have a good look at how I've changed the routes. Your old work is still in here but will need relinking in your ticket.


stevewalton28 (approved):

app/assess/templates/assesser_project_view.html
```diff
@@ -0,0 +1,43 @@
...
+<span class="govuk-caption-l">Project reference</span>
+<h1 class="govuk-heading-l">{{application_data.identifier}}</h1>
+
+    {{ govukDetails({
+        'summaryText': fund_data.name,
+        'text': fund_data.description,
+        'id': "assessment-page-details",
+        'open':True}) }}
+
+        <script nonce="{{ csp_nonce() }}">
```
harryyo, line 24: Nice nonce! One thought though - as this is an accessibility feature, we'll probably need it on all govukDetails components right? So perhaps could this script be more global and go in the footer, perhaps in a document.onLoad or similar and auto-close all elements with the govuk-details classname maybe?

app/templates/head.html
```diff
@@ -3,3 +3,10 @@
 <meta name="author" content="{{ service_meta_author }}">
 <!--[if gt IE 8]><!--><link rel="stylesheet" type="text/css" href="{{ url_for('static', filename='govuk-frontend-4.0.0.min.css') }}" /><!--<![endif]-->
 <!--[if IE 8]><link rel="stylesheet" type="text/css" href="{{ url_for('static', filename='govuk-frontend-ie8-4.0.0.min.css') }}" /><![endif]-->
+<noscript>
+    <style>
+        .noscriptHide {
```
harryyo, line 8: Where abouts is this used? (I haven't managed to find it...!)

app/assess/routes.py
```diff
@@ -72,3 +72,30 @@ def application(fund_id: str, round_id: str, application_id: str):
...
+        abort(404)
+
+    application_data = get_application(
+        fund_id=fund_id, identifier=application_id
+    )
+    if not application_data:
+        abort(404)
+
+    return render_template(
+        "assesser_project_view.html",
```
harryyo, line 97: Could this template nam be simplified to match the route name a bit more closely? So it's easy to identify which templates go with which routes when we have lots of them in future? - appreciate there's a conflict with the legacy view of last sprint, but could something like "project_summary" work?

runtime.txt
```diff
@@ -1 +1 @@
-python-3.10.1
+python-3.10.x
```
harryyo, line 1: 👍

app/assess/templates/assesser_project_view.html
SentientHousePlant: Great spot Harry! I will move it into the head.

app/create_app.py
```diff
@@ -48,7 +48,10 @@ def create_app() -> Flask:
 
     Compress(flask_app)
     Talisman(
-        flask_app, content_security_policy=csp, strict_transport_security=hss
+        flask_app,
+        content_security_policy=csp,
+        strict_transport_security=hss,
+        content_security_policy_nonce_in=["script-src"],
```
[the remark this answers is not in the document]
SentientHousePlant: You're right! This is an artefact from the template we used.

runtime.txt
SentientHousePlant: Thank you 💯

harryyo (approved):

PR #11, opened 2022-04-08
FS-727 template question view

Adam-W1:
FS-727: Implement standard template for assessment view of question (without scoring)


app/assess/data.py
```diff
@@ -58,6 +50,15 @@ def get_local_data(endpoint: str):
         return api_data.get(endpoint)
 
 
+def call_search_applications(params: dict):
+    applications_endpoint = (
+        APPLICATION_STORE_API_HOST
+        + APPLICATIONS_SEARCH_ENDPOINT.format(params=urlencode(params))
```
SentientHousePlant, line 56: I haven't seen this before but it would actually be really useful to use. I'll try and use url encode for url args from now :) good one!

app/assess/data.py
```diff
@@ -58,6 +50,15 @@ def get_local_data(endpoint: str):
         return api_data.get(endpoint)
 
 
+def call_search_applications(params: dict):
```
SentientHousePlant, line 53: Docstrings? :)

app/assess/routes.py
```diff
@@ -1,10 +1,10 @@
-from app.assess.data import APPLICATION_SEARCH_ENDPOINT
+from app.assess.data import APPLICATIONS_SEARCH_ENDPOINT
 from app.assess.data import get_application
 from app.assess.data import get_applications
 from app.assess.data import get_fund
 from app.assess.data import get_funds
 from app.assess.data import get_questions
-from app.assess.data import get_round
+from app.assess.data import get_round_with_applications
```
SentientHousePlant, line 7: Not sure what the convention is when you're importing so much from a single file...I think using 
```python
from app.assess.data import *
```
Might be worth doing at this point.

app/assess/templates/components/answer_view_table.html
```diff
@@ -0,0 +1,23 @@
+{% macro render(question, question_search_endpoint) -%}
+<table class="govuk-table">
```
SentientHousePlant, line 2: nice macro for a table :) this is different from the default govuk table in some way to warrant its existence?

SentientHousePlant (changes requested):
Very nice! Just some mild comments but overall very nice code - you're becoming a python pro already!! 🔥 💯

app/assess/routes.py
Adam-W1: good shout - changed

harryyo (approved):
I think all nice!

app/assess/data.py
harryyo: Nice refactor

app/assess/routes.py
harryyo: Yep good idea

app/assess/routes.py
```diff
@@ -68,15 +65,50 @@ def landing():
...
+
+    application = get_application_status(application_id=application_id)
+    if not application:
+        abort(404)
+
+    fund = get_fund(application.fund_id)
+    if not fund:
+        abort(404)
+
+    # application_summary = get_application_summary()
```
harryyo, line 96: Oh these can be removed now

app/assess/templates/components/answer_view_table.html
Adam-W1: only that its html and i prefer looping with the html than the nunjuck macro. The macros are not within the scope of this ticket but its a fair point - why reinvent.  So i will look into that 💯

stevewalton28 (approved):

harryyo (approved):

PR #16, opened 2022-06-24
FS 947 template text input

ramsharma-prog:
1. Reusable **HTML** template for text input.
2. **import macro** to use the template :      {% from "components/text_input_template.html" import textInput as **text** %}.
3.  insert value to be displayed such as **`JSON`** contains key **value** pair:    {{  text (_**value**_)  }}.


app/assess/models/question_field.py
```diff
@@ -16,3 +16,9 @@ def from_json(data: dict):
             field_type=data.get("field_type"),
             answer=data.get("answer"),
         )
+
+    @staticmethod
+    def get_question_fields(json):
+        for pages in json["questions"]:
+            for page in pages["fields"]:
+                return page
```
harryyo, line 24: I don't think you'd really need this method because a QuestionField() instance already has title and answer properties so you can just use those like question.title, question.answer

app/assess/routes.py
```diff
@@ -138,3 +139,49 @@ def fund_round(fund_id: str, round_id: str):
...
+                        "type": "text",
+                        "answer": "99 evoco, example street, London, UB5 5FF",
+                    },
+                ],
+            }
+        ],
+    }
+
+    name = QuestionField.get_question_fields(input_text_name)
+    address = QuestionField.get_question_fields(input_text_address)
```
harryyo, line 181: Instead of this section I think you could just do 
name = QuestionField(key="abc",title="Your name",field_type="text",answer="Steve")
address = QuestionField(key="xyz",title="Your UK address",field_type="text",answer="99 evoco, example street, London, UB5 5FF")
As there's no need to reprocess json data with a separate function (get_question_fields) as from_json already does that

harryyo (commented):
Hi Ram, great, just small comments

app/assess/models/question_field.py
ramsharma-prog: okay, so I used the from_json function instead 😄

app/assess/routes.py
ramsharma-prog: I used the from_json function instead 😄

app/assess/templates/macros/example_text_input.html
```diff
@@ -0,0 +1,10 @@
+{% extends "base.html" %}
+{% from "macros/text_input.html" import textInput as text %}
+
+{% block content %}
+
+{{ text(name.title, name.answer) }}
+{{ text(address.title, address.answer)}}
+
+
+{% endblock %}
```
gio-karunakaran, line 10: 💯 great use of macros :D

app/assess/routes.py
```diff
@@ -138,3 +140,53 @@ def fund_round(fund_id: str, round_id: str):
...
+            }
+        ],
+    }
+
+    name = QuestionField.from_json(
+        input_text_name["questions"][0]["fields"][0]
+    )
+    address = QuestionField.from_json(
+        input_text_address["questions"][0]["fields"][0]
+    )
```
gio-karunakaran, line 186: 👍 love the use of data classes here

gio-karunakaran (approved):
LGTM! 🙌

Adam-W1 (approved):
Nice one - Ill rebase onto this once merged and apply the other macro 👍

PR #17, opened 2022-06-28
Fs 946 template structured question

Adam-W1:
A macro to display formatted/structured questions for Assessors.


gio-karunakaran (commented):
all good! just one thought/ query :)

app/assess/templates/fragments_base.html
```diff
@@ -0,0 +1,11 @@
+{% extends "base.html" %}
+{% from "components/structured_question_fragment.html" import structuredQuestion as structured_question %}
+
+{% block content %}
+
+<!-- handle selection type data here by passing it into structured_question template -->
+{% for key, values in data["select_type_data"].items() %}
+    {{ structured_question(key, values) }}
+{% endfor %}
```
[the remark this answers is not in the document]
Adam-W1: good point ill take a look at good practice

app/assess/templates/macros/structured_question_macro.html
```diff
@@ -1,7 +1,8 @@
 {% macro structuredQuestion(select_type_data) %}
-<!-- select_type_data is in the format: key -> string, values -> list -->
 
-{% for key, values in select_type_data.items() %}
+<h2 class="govuk-heading-l">{{select_type_data["title"]}}</h3>
+
+{% for key, values in select_type_data["answers_per_question"].items() %}
 <h3 class="govuk-body govuk-!-font-weight-bold">
```
SentientHousePlant, line 6: 🔧 Can we not used a gov-uk-heading style instead here?

SentientHousePlant (changes requested):
Just one small note for you :)

app/assess/templates/macros/structured_question_macro.html
Adam-W1: good spot 🔍

harryyo (approved):
Very good, only one comment

app/assess/models/structured_question.py
```diff
@@ -0,0 +1,45 @@
+import ast
+from dataclasses import dataclass
+from typing import Dict
+from typing import List
+
+from .question import Question
+
+
+@dataclass
+class StructuredQuestionView:
```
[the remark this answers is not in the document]
Adam-W1: Im not sure actually, with a fresh head, im thinking this should be a instance method on the Question class, as we are returning a question in a type of format (structured question type)

app/assess/models/question.py
```diff
@@ -26,3 +27,20 @@ def add_field(self, field: QuestionField):
...
+            answer = question_field.answer
+            try:
+                if type(ast.literal_eval(answer)) == list:
+                    answers_per_question_field[
+                        question_field.title
+                    ] = ast.literal_eval(answer)
+                    continue
+            except (ValueError, SyntaxError):
+                # literal_eval cannot parse string values
+                pass
```
SentientHousePlant, line 43: 🤔 perhaps log something here or try parsing it another way? Seems dodgy to have a except-pass block in your code.

SentientHousePlant (changes requested):

app/assess/models/question.py
Adam-W1: Ok, I have read using exceptions for flow control is quite normal in Python (as opposed to some other languages), the alternative was a bit higher on the cognitive complexity scale. Im happy to avoid certain flow controls if thats an approach we want to use on the project though 👀

SentientHousePlant (approved):

PR #19, opened 2022-06-29
📦 Added pip-compile functionality.

SentientHousePlant:
That's right 🎉 It now uses pip-compile, 21st package management.

# But how do I use pip-compile?

✨ ✨  The `.in` files contain handpicked, bespoke dependencies of the project. **_We_** have to add them in. When we add new deps we run `pip-compile` to turn these into the lock-files - that is the standard `requirements.txt` and `dev-requirements.txt`


.github/workflows/govcloud.yml
```diff
@@ -23,20 +23,13 @@ jobs:
...
-      - name: Install Chromium
-        uses: browser-actions/setup-chrome@latest
-      - name: setup Chromedriver
-        uses: nanasess/setup-chromedriver@v1
-        with:
-        # Optional: do not specify to match Chrome's version
-          chromedriver-version: '88.0.4324.96'
       - name: Run Accessability tests
-        run: pytest --driver Chrome --driver-path .venv/lib/python3.10/site-packages/chromedriver_py/chromedriver_linux64
+        run: pytest
```
gidsg, line 32: Seems like we're running all tests already above in https://github.com/communitiesuk/funding-service-design-assessment/blob/292e37302cead998cb13001ac637b23ccb0dc8d3/.github/workflows/govcloud.yml#L30 so don't think this is adding any value?

tests/conftest.py
```diff
@@ -63,15 +63,16 @@ def flask_test_client():
 
 
 @pytest.fixture(scope="class")
-def selenium_chrome_driver(request):
+def selenium_chrome_driver(request, live_server):
```
gidsg, line 66: Looks like `live_server` is unused could it be removed?

harryyo (approved):
Great

Adam-W1 (approved):
🚀

PR #21, opened 2022-07-06
FS-1009 Assets Build Fix

harryyo:
### Assets Build Fix
- Update to assets loading and static_folder configuration, to keep config in config
- Define STATIC_FOLDER and STATIC_URL_PATH in default config **only**
- Give more descriptive warning in the event of a build error

### DIFF
--- a/app/assets.py
+++ b/app/assets.py
@@ -1,9 +1,10 @@
 """Compile static assets."""
 from config import Config
 from flask_assets import Bundle
+from webassets.exceptions import BuildError
 
 
-def compile_static_assets(assets):
+def compile_static_assets(assets, flask_app):
     """Configure and build asset bundles."""
 
     # Main asset bundles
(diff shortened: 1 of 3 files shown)


app/assets.py
```diff
@@ -1,9 +1,9 @@
...
-from config import Config
 from flask_assets import Bundle
+from webassets.exceptions import BuildError
 
 
-def compile_static_assets(assets):
+def compile_static_assets(assets, flask_app):
     """Configure and build asset bundles."""
 
     # Main asset bundles
```
Adam-W1, line 9: i think it may be worth adding into the comment that flask static folder is set in the new Env var: STATIC_URL_PATH. what do you think?

Adam-W1 (commented):
Really nice solution 👍 just a couple comments

ramsharma-prog:
LGTM

ramsharma-prog (approved):

app/assets.py
harryyo: Good point, I've beefed up the docstring in the compile_static_assets function to describe how it's put together more fully as isn't obvious.

Adam-W1 (approved):
🔥

PR #32, opened 2022-10-05
Fs 1530 enable dependabot

Adam-W1:
https://digital.dclg.gov.uk/jira/browse/FS-1530


srh-sloan (approved):
👍

gidsg:
@Adam-W1 can we remove the `safety` workflow check at the same time?

PR #44, opened 2022-10-06
Bau check for null

srh-sloan:
Added null check around date_submitted from application

### DIFF
--- a/config/envs/development.py
+++ b/config/envs/development.py
@@ -4,8 +4,8 @@ from fsd_utils import configclass
 
 @configclass
 class DevelopmentConfig(DefaultConfig):
-    ASSESSMENT_STORE_API_HOST = "https://funding-service-design-assessment-store-dev.london.cloudapps.digital"  # noqa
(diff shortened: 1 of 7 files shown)


config/envs/development.py
```diff
@@ -4,8 +4,8 @@
 
 @configclass
 class DevelopmentConfig(DefaultConfig):
-    ASSESSMENT_STORE_API_HOST = "https://funding-service-design-assessment-store-dev.london.cloudapps.digital"  # noqa
+    # ASSESSMENT_STORE_API_HOST = "https://funding-service-design-assessment-store-dev.london.cloudapps.digital"  # noqa
```
harryyo, line 7: Can this go altogether?

harryyo (changes requested):
Great, just a comment

config/envs/development.py
srh-sloan: Yes, good spot!

harryyo (approved):

PR #51, opened 2022-11-10
added api URLs to test manifest

srh-sloan:
Assessment was borken on test after last merge as the test manifest didn't contain the URLs for fund_store or assessment_store. This adds them


gidsg (approved):
👍

PR #52, opened 2022-11-15
Fs 1864 - Search Ref and project name

adamdavies1:
Can now use a search bar to filter through the application overviews
image


app/assess/templates/components/application_overviews_table.html
```diff
@@ -1,10 +1,30 @@
...
+            </svg>
+          </button>
+        </div>
+      </div>
+    </form>
+  </div>
+
+</div>
+
+<table class="govuk-table" data-module="sortable-table", id="application_overviews_table">
```
srh-sloan, line 26: Do we still need the `sortable-table` bit?

srh-sloan (approved):
1 tiny comment - looks good!

app/assess/routes.py
```diff
@@ -294,8 +295,17 @@ def landing():
     of applications and their statuses
     """
 
+    search_params = {
+        "search_term": "",
+    }
```
gidsg, line 300: Do we want to pass this through as empty if not specified? Can we make it `None` in that case?

adamdavies1: No as it will add None to the search bar
image

app/assess/templates/components/application_overviews_table.html
adamdavies1: yup removed this

app/assess/routes.py
gidsg: @adamdavies1 I think we could solve this by adding a condition to the template e.g.

`{{ query.search_term if query.search_term }}`

PR #53, opened 2022-11-16
Fs 1868 filter by asset and status

srh-sloan:
Adds filter by asset and/or status to the assessor dashboard.

Includes FS-1868 and FS-699

### DIFF
--- a/app/assess/routes.py
+++ b/app/assess/routes.py
@@ -1,6 +1,7 @@
 from app.assess.data import *
 from app.assess.data import get_application_overviews
 from app.assess.data import submit_score_and_justification
+from app.assess.display_value_mappings import assessment_statuses
 from app.assess.display_value_mappings import asset_types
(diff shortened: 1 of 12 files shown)


adamdavies1 (approved):
LGTM

app/assess/routes.py
```diff
@@ -297,6 +299,9 @@ def landing():
 
     search_params = {
         "search_term": "",
+        "search_in": "project_name,application_short_id",
```
gidsg, line 302: Just wondering whether this should be a store concern, rather than letting the frontend decide which fields to query?

srh-sloan: We thought a frontend thing as the front end will be telling users which fields the search applies to, so rather than duplicate that info it's all in one service.

The frontend search box has the label 'Search in project name and ref' or similar.

PR #55, opened 2022-11-17
Add clear filters

srh-sloan:
Added clear filters button as per current prototype and discussions with Emma.

It only appears when searching/filtering is applied.

https://fsd-pre-award.herokuapp.com/s26-assessment/search-filtering


adamdavies1 (approved):
LGTM, works great!

PR #58, opened 2022-11-24
FS-1581: Add templates for Assessor Task List page

tferns:
**Summary:**

- Add templates outlined by jira story [FS-1581](https://digital.dclg.gov.uk/jira/browse/FS-1581)
- Consumes mapping configuration from:
    - https://github.com/communitiesuk/funding-service-design-assessment-store/pull/50
- Add test for new section/criteria macro components

**Considerations:**
- Might be a better way to pull back the fund_name? As opposed to another API call?
- Styling may change (future work?) - as new prototype is in "Considerations" phase at the time of this edit.


Adam-W1:
> * FS-1581

Just on the fourth point, the config will be consumed from the assessment store

tferns:
> > * FS-1581
> 
> Just on the fourth point, the config will be consumed from the assessment store

Thanks, have updated the description to reflect that just now

app/assess/templates/macros/scored_section.html
```diff
@@ -0,0 +1,52 @@
+{%- from "govuk_frontend_jinja/components/table/macro.html" import govukTable -%}
```
Adam-W1, line 1: A bit of a weird one to wrap the head around, but technically we don't have scored sections. We have unscored sections and scored criteria, like so:
scored_criteria -> sub_criteria -> theme -> answers
unscored_sections -> sub_criteria -> theme -> answers

tferns: Hey, thanks for clarifying that for me this morning.  It's a lot more clear now...

 I've pushed https://github.com/communitiesuk/funding-service-design-assessment/pull/58/commits/eb884c180a63acf4e781fbfbbc8eba0d863659c6 to implement the changes we spoke about, if you have any other suggestions please let me know! 👍

app/assess/models/assessor_task_list.py
```diff
@@ -0,0 +1,85 @@
...
+@dataclass
+class _Criteria:
+    name: str
+    total_criteria_score: int
+    total_criteria_score_possible: int
+    weighting: float
+    sub_criterias: List[_CriteriaSubCriteria]
+
+
+@dataclass
```
Adam-W1, line 33: very neat use of data classes here 🔥

config/envs/default.py
```diff
@@ -52,15 +52,21 @@ class DefaultConfig:
...
 
+    APPLICATION_OVERVIEW_ENDPOINT_APPLICATION_ID = (
+        "/application_overviews/{application_id}"
+    )
+
     ASSESSMENT_SCORES_ENDPOINT = (
         ASSESSMENT_STORE_API_HOST
         + "/assessments/{assessment_id}/sub_criterias/{sub_criteria_id}/scores"
     )
+    ASSESSMENT_MAPPING_CONFIG_ENDPOINT = (
```
Adam-W1, line 67: im not sure this is used?

app/assess/models/assessor_task_list.py
```diff
@@ -0,0 +1,85 @@
...
+                    total_criteria_score=criteria["total_criteria_score"],
+                    total_criteria_score_possible=criteria[
+                        "total_criteria_score_possible"
+                    ],
+                    weighting=criteria["weighting"],
+                    sub_criterias=[
+                        _CriteriaSubCriteria(
+                            id=sub_criteria["id"],
+                            name=sub_criteria["name"],
+                            status=sub_criteria["status"],
```
Adam-W1, line 76: I think we are reading the mapping config here which doesnt contain any state. We wont be saving scores and statuses to the config dict, they should be in the assessment database. 
The statuses and scoring will come from the db records associated with the criteria/sub_critera. I think this Is linked to my comment on the assessment store PR, the state needs to be matched there before sending over to assessment.
so Assessment store reads (and matches) the config to the db records and then returns the sections and the ordered criteria with associated criteria score and sub_criteria statuses
Happy to chat this through on a call

config/envs/default.py
tferns: Good spot, I'll push a commit to remove it.  Left over from the old PR...

app/assess/models/assessor_task_list.py
tferns: > I think we are reading the mapping config here which doesnt contain any state

The structure is similar to the mapping config, however there are subtle differences, the root level dataclass `AssessorTaskList` is meant to represent a model which has only the state necessary to be rendered on the UI.

State can be added as part of this transformation:  [assessment_mapping_fund_round.py#L30-L39](https://github.com/communitiesuk/funding-service-design-assessment-store/blob/cf75286e37b0cfe1f95423e8aea0ae793ec6502d/config/mappings/assessment_mapping_fund_round.py#L30-L39)

We will have to query the database for this state though.  

> the state needs to be matched there before sending over to assessment

This is what's happening in the above assessment_store function, however there aren't any database queries to get the state and then transform it, there are just placeholders at the moment -- I think I could build out the queries based on the current structure/data model & doing TDD, but I was reluctant to as they might have changes from the work Gio is doing at the moment on scores (also wanted to get feedback on approach so far before continuing).

app/assess/models/assessor_task_list.py
```diff
@@ -0,0 +1,85 @@
+from dataclasses import dataclass
+from typing import List
+
+
+@dataclass
+class _SubCriteria:
```
Adam-W1, line 6: I imagine this would be separated out as the dataclass will be used elsewhere. IMO okay to leave here for this PR though and refactor as needed

Adam-W1 (approved):
looks good 👍

tests/conftest.py
```diff
@@ -7,7 +8,8 @@
 from selenium.webdriver.chrome.service import Service
 from webdriver_manager.chrome import ChromeDriverManager
 
-multiprocessing.set_start_method("fork")  # Required on macOSX
+if platform.system() == "Darwin":
+    multiprocessing.set_start_method("fork")  # Required on macOSX
```
SentientHousePlant, line 12: 💯

SentientHousePlant (approved):
Love work! Comprehensive tests and clean code + seperation of concerns. I like the use of filters to avoid pre-processing data before passing to routes.

PR #59, opened 2022-11-29
Fs 1580 score app sub crit

gio-karunakaran:


app/assess/data.py
```diff
@@ -128,20 +132,38 @@ def get_round_with_applications(
     return None
 
 
+def get_score_and_justification(application_id, sub_criteria_id):
+    url = Config.ASSESSMENT_SCORES_ENDPOINT
+    params = {
+        "application_id": application_id,
+        "sub_criteria_id": sub_criteria_id,
+    }
+    response = requests.get(url, params)
```
gidsg, line 141: could we use `get_data` for this as above?

app/assess/data.py
```diff
@@ -128,20 +132,38 @@ def get_round_with_applications(
...
+    # current_app.logger.info(
+    #     "Response from Assessment Store:" + response.content
+    # )
+    if response.status_code == 200:
+        return response.json()
+    else:
+        return None
```
gidsg, line 148: Can we raise an exception, or at least log at error level in the non 200 case? This will make the issue less opaque if the request fails. I think ideally `get_data` should do the same.

app/assess/routes.py
```diff
@@ -30,12 +34,61 @@ def funds():
...
+        except AttributeError:
+            user_id = (  # TODO remove and force g.account_id after adding authentication # noqa
+                ""
+            )
+        submit_score_and_justification(
+            score=score,
+            justification=justification,
+            application_id=application_id,
+            user_id=user_id,
+            timestamp=datetime.now().strftime("%Y/%m/%d %H:%M:%S"),
```
gidsg, line 63: Times should be in UTC and ISO format, so I think  `datetime.utcnow().isoformat()` is what you want here, if you can test if that works with the assessment-store?

app/assess/templates/sub_criteria.html
```diff
@@ -0,0 +1,105 @@
...
+                        <p class="govuk-body-s">There are currently no comments&nbsp;for this section.
+                            <!--<a href="#" class="govuk-link govuk-link--no-visited-state">Add a comment</a>-->
+                        </p>
+
+                    </section>
+                    <section class="comment-section">
+                        <h3 class="govuk-heading-m">Local support</h3>
+
+                        <p class="govuk-body-s">There are currently no comments&nbsp;for this section.
+                            <!--<a href="#" class="govuk-link govuk-link--no-visited-state">Add a comment</a>-->
```
gidsg, line 84: We can use Jinja comments instead of HTML comment syntax? https://jinja.palletsprojects.com/en/3.1.x/templates/#comments

app/assess/data.py
```diff
@@ -74,12 +74,16 @@ def get_fund(fund_id: str) -> Union[Fund, None]:
...
+    try:
+        if len(response) > 0:
+            fund = Fund.from_json(response[0])
+            if "rounds" in response and len(response["rounds"]) > 0:
+                for fund_round in response["rounds"]:
+                    fund.add_round(Round.from_json(fund_round))
+            return fund
+    except KeyError:
+        if type(response) == dict:
+            return response
```
tferns, line 86: This seems weird that you have to manage two different types from an endpoint

tferns: Seems like this endpoint only returns a dict - maybe we can remove the array specific logic?

https://github.com/communitiesuk/funding-service-design-fund-store/blob/225163028d60335102570850c212ffb8b0ab20a7/core/funds.py#L72

Might be good to add a test for this

app/assess/routes.py
```diff
@@ -30,12 +34,71 @@ def funds():
...
 
 
+@login_requested
+@assess_bp.route(
+    "score/application_id/<application_id>/sub_criteria_id/<sub_criteria_id>",
+    methods=["POST", "GET"],
+)
+def application_sub_crit_scoring(application_id: str, sub_criteria_id: str):
+    fund = get_fund(Config.COF_FUND_ID)
+    form = JustScoreForm()
```
tferns, line 48: If we renamed `JustScore` - it might be good to also rename this form, minor though

app/assess/routes.py
```diff
@@ -30,12 +34,71 @@ def funds():
...
+            except AttributeError:
+                user_id = (  # TODO remove and force g.account_id after adding authentication # noqa
+                    ""
+                )
+            submit_score_and_justification(
+                score=score,
+                justification=justification,
+                application_id=application_id,
+                user_id=user_id,
+                timestamp=datetime.now().strftime("%Y/%m/%d %H:%M:%S"),
```
tferns, line 67: I don't think timestamp is required, since it's being handled by the database

app/assess/data.py
```diff
@@ -128,20 +132,38 @@ def get_round_with_applications(
...
-    assessment_id, person_id, score, justification, sub_crit_id
+    score, justification, application_id, user_id, timestamp, sub_criteria_id
 ):
-
     data_dict = {
         "score": score,
         "justification": justification,
-        "person_id": person_id,
+        "user_id": user_id,
+        "timestamp": timestamp,
```
tferns, line 158: Timestamp is unnecessary

app/assess/routes.py
```diff
@@ -30,12 +34,71 @@ def funds():
...
+            scores_submitted = True
+            score_error, justification_error = False, False
+
+        else:
+            scores_submitted = False
+            score_error = True if not form.score.data else False
+            justification_error = True if not form.justification.data else False
+
+    # call to assessment store to get the latest score if request = get
+    latest_score = get_score_and_justification(application_id, sub_criteria_id)
```
tferns, line 79: This will be executed on both `POST` and `GET` requests, so the comment is a little deceiving 

I was thinking there might be a race condition here... but I think `submit_score_and_justification` will only return once the request has succeeded & by that time the database insertions will have happened on the store

app/assess/routes.py
```diff
@@ -30,12 +34,71 @@ def funds():
...
+    # TODO make COF_score_list extendable to other funds
+    COF_score_list = [
+        (5, "Strong"),
+        (4, "Good"),
+        (3, "Satisfactory"),
+        (2, "Partial"),
+        (1, "Poor"),
+    ]
```
tferns, line 87: Out of interest, by extendable do you mean configurable?  As in funds (in the future) should be able to create their own scoring criteria?

app/assess/routes.py
```diff
@@ -30,12 +34,71 @@ def funds():
...
+            submit_score_and_justification(
+                score=score,
+                justification=justification,
+                application_id=application_id,
+                user_id=user_id,
+                timestamp=datetime.now().strftime("%Y/%m/%d %H:%M:%S"),
+                sub_criteria_id=sub_criteria_id,
+            )
+            scores_submitted = True
+            score_error, justification_error = False, False
```
tferns, line 71: These variables are already defined as `False` on L50

config/envs/default.py
```diff
@@ -70,7 +67,7 @@ class DefaultConfig:
 
     TALISMAN_SETTINGS = CommonConfig.TALISMAN_SETTINGS
 
-    COF_FUND_ID = CommonConfig.COF_FUND_ID
-    COF_ROUND2_ID = CommonConfig.COF_ROUND_2_ID
+    COF_FUND_ID = "47aef2f5-3fcb-4d45-acb5-f0152b5f03c4"
+    COF_ROUND2_ID = "c603d114-5364-4474-a0c4-c41cbf4d3bbd"
```
tferns, line 71: Is there a reason for this change?

config/envs/development.py
```diff
@@ -5,14 +5,20 @@
...
+    # FUND_STORE_API_HOST = "https://funding-service-design-fund-store-dev.london.cloudapps.digital" # noqa
+    # ASSESSMENT_STORE_API_HOST = "https://funding-service-design-assessment-store-dev.london.cloudapps.digital" # noqa
+
+    # for local dev/test use with flask run and USE_LOCAL_DATA = False:
+    FUND_STORE_API_HOST = (  # change in fund store repo and adjust here accordingly # noqa
+        "http://127.0.0.1:5002"
+    )
+    ASSESSMENT_STORE_API_HOST = (  # change in assessment store repo and adjust here accordingly # noqa
+        "http://127.0.0.1:5001"
+    )
```
tferns, line 23: Is this is specific to your own local environment I don't know if it's a good idea to commit it to the repo, might break some peoples local environments if they have those apps running on slightly different ports or only accessible with a `host.docker.internal` address etc.

app/assess/templates/sub_criteria.html
```diff
@@ -0,0 +1,105 @@
...
+                </div>
+            </div>
+
+            <div class="govuk-grid-column-full">
+
+                <div class="govuk-grid-column-one-third">
+
+                    <nav class="assessment-navigation">
+                        <ul class="govuk-list">
+                            <li><a class="govuk-link--no-visited-state"
```
tferns, line 56: Is the `no-visited-state` on purpose or hangover from the prototype?

app/assess/templates/macros/scores.html
```diff
@@ -1,11 +1,26 @@
 {%- from 'govuk_frontend_jinja/components/radios/macro.html' import govukRadios -%}
 
-{% macro scores(score_radio_id) %}
+{% macro scores(score_form_name, score_tuple, score_error) %}
```
tferns, line 3: `score_tuple` is a list of `tuples`

app/assess/templates/macros/scores.html
```diff
@@ -1,11 +1,26 @@
...
+        <div class="govuk-radios" data-module="govuk-radios">
+            {% for i in score_tuple %}
+            <div class="govuk-radios__item">
+                <input class="govuk-radios__input" id={{"score-{}".format(i[0])}} name={{score_form_name}}
+                    type="radio" value={{i[0]}}>
+                <label class="govuk-label govuk-radios__label" for={{score_form_name}}>
+                    <span class="govuk-!-font-weight-bold">{{i[0]}}</span> {{i[1]}}
+                </label>
+            </div>
+            {% endfor %}
```
tferns, line 21: Can we destructure each tuple so we have more clear/declarative naming rather than indexing?
```suggestion
            {% for score, description in score_tuple %}
            <div class="govuk-radios__item">
                <input class="govuk-radios__input" id={{"score-{}".format(score)}} name={{score_form_name}}
                    type="radio" value={{score}}>
                <label class="govuk-label govuk-radios__label" for={{score_form_name}}>
                    <span class="govuk-!-font-weight-bold">{{score}}</span> {{description}}
                </label>
            </div>
            {% endfor %}
```

tferns (changes requested):
Good work on this, have left a few suggestions throughout the code

app/assess/data.py
gio-karunakaran: using get data now 👍🏾

app/assess/data.py
gio-karunakaran: have moved this logic over to get-data and we are now logging failed requests in there 👍🏾

app/assess/routes.py
gio-karunakaran: this is getting auto-created from assessment store now so will remove this from here completely  👍🏾

app/assess/data.py
gio-karunakaran: have removed 👍🏾

app/assess/data.py
```diff
@@ -13,18 +13,26 @@
...
+            current_app.logger.info(f"Fetching data from '{endpoint}'.")
+            response = requests.get(endpoint)
         if response.status_code == 200:
-            data = response.json()
+            return response.json()
         else:
+            current_app.logger.error(
+                f"Could not get data for endpoint '{endpoint}' "
+            )
             return None
```
tferns, line 35: Wonder if it would be better to actually raise an exception that bubbles to the top so people know that's actually what's went wrong as opposed to when `None` is returned and they try to access it, just food for thought though not specific to your PR

app/assess/data.py
```diff
@@ -128,20 +138,34 @@ def get_round_with_applications(
...
+        "sub_criteria_id": sub_criteria_id,
     }
-    url = Config.ASSESSMENT_SCORES_ENDPOINT.format(
-        assessment_id=assessment_id, sub_criteria_id=sub_crit_id
-    )
+    url = Config.ASSESSMENT_SCORES_ENDPOINT
     response = requests.post(url, json=data_dict)
-    print(response.content)
+    current_app.logger.info(
+        f"Response from Assessment Store: '{response.json}'."
```
tferns, line 167: `json` is a method not a property

```suggestion
        f"Response from Assessment Store: '{response.json()}'."
```

tests/conftest.py
```diff
@@ -1,13 +1,11 @@
...
-
 import pytest
 from app.create_app import create_app
 from selenium import webdriver
 from selenium.webdriver.chrome.options import Options
 from selenium.webdriver.chrome.service import Service
 from webdriver_manager.chrome import ChromeDriverManager
 
-multiprocessing.set_start_method("fork")  # Required on macOSX
+# multiprocessing.set_start_method("fork")  # Required on macOSX
```
tferns, line 8: Might be good to do a conditional here rather than commenting it out

Maybe 

```python3
if platform.system() == "Darwin":
    multiprocessing.set_start_method("fork")  # Required on macOSX
```

tferns (approved):
Thanks for addressing my comments, approving for the things I commented on

PR #63, opened 2022-12-06
fs-1968-score-history

gio-karunakaran:


app/assess/routes.py
```diff
@@ -87,6 +90,7 @@ def application_sub_crit_scoring(application_id: str, sub_criteria_id: str):
         "sub_criteria.html",
         scores_submitted=scores_submitted,
         form=form,
+        score_list=score_list if len(score_list) > 0 else None,
```
tferns, line 93: Would this be better defaulted to an empty collection?
```suggestion
        score_list=score_list if len(score_list) > 0 else [],
```

app/assess/templates/macros/scores.html
```diff
@@ -10,7 +10,7 @@
           </p>
         {% endif %}
         <div class="govuk-radios" data-module="govuk-radios">
-            {% for score, description in score_tuple %}
+            {% for score, description in score_list %}
```
tferns, line 13: My worry is if `score_list` is `None` then we'll get a `TypeError` but not sure if it being a jinja2 template guards against that, if it's an empty collection it'll just skip over it.

gio-karunakaran: I'm managing this by doing an if score_list block in the jinja template, and as the logic is pythonic None is falsey so will just not display the score history stuff rather than errorring out which is ideal behavior!

PR #64, opened 2022-12-07
BAU - updating styling to match prototype

stevewalton28:
Updating styling to match prototype:

- When I picked this up, working against local data was fine, but wasn't integrated/wasn't working against real APIs. Have made some adjustments to the `data.py` code to make sure that this works against the real APIs for fund store
- Added some styling changes to the dashboard and to the assessor task list page, which included adding some additional fields to the state for the assessment dashboard


tferns (approved):
Looks good but the builds are failing, you can change to use `3.10.x` if you want to make sure the tests pass before merging.

PR #69, opened 2022-12-13
Fs 1972 run shared tests

srh-sloan:
Adding step to workflow to call `run-shared-tests.yml` from the shared workflows


gidsg (approved):

gidsg:
Would be good to squash and merge as commit history is quite messy.

PR #73, opened 2022-12-15
FS-1987: Front-end for sub criteria theme answers & more

tferns:
This is a pretty hefty change.  I'll try to section the description into smaller chunks to describe each change.

### Consuming https://github.com/communitiesuk/funding-service-design-assessment-store/pull/58

This endpoint is now being called whenever the `display_sub_criteria` is called, given a `theme_id` is provided.
There has been mock/local data added for testing this component without full integration with the store.

If you wish to use this, you will need to make sure all relevant API calls are using the application identifier `app_123` and occasionally `1a2b3c4d` as a sub_criteria_id if the situation calls for it.  

### Pre-processing after consumption [app/assess/models/ui/applicators_response.py:314](https://github.com/communitiesuk/funding-service-design-assessment/blob/103a3171e3182de289ae516b16d4e2cca9033c6a/app/assess/models/ui/applicators_response.py#L314)

After we consume the data, we do some transformations on it to make it easier to work with for feeding into the UI components, the three main things that happen are:

#### Transformation 1: `heading` `description` `amount`
Some data comes through split between 3 items -  "heading", "description", "amount", these are transformed into a single "grouped_fields" item,  we join them all by their field id and parse them into a single item which can easily be processed by the existing ui components, in all known scenarios of this they eventually get fed into `MonetaryKeyValues` as they display financial data.

#### Transformation 2: `checkbox`
Checkbox items come through as a single item, the way that this works is that the answer is populated with a list of dash separated keys, these represent all options the user **has** selected, options a user hasn't selected do not come through.  We transform these into regular **text** items with an answer of "Yes", and do so in the order they come in, this allows the UI to handle them as `BesideQuestionAnswerPair` to display the question beside a "Yes"

#### Transformation 3: `non-number grouped_fields`
When non-number specific `grouped_fields` come in, they typically contain multiple questions & answers that need to be displayed on the UI.  Similar to checkbox - we parse these and create regular **text** items, so they can be displayed on the UI using the existing components.  We identify non-number fields by assuming any item without a `field_type` of `numberField` is not a number.

#### Sorting
During each transformation step, we keep track of the unique `field_ids` which have been processed, these are used to find any items which have not been processed.  We create a new list of `post_processed_items` constructed from the three transformations & unprocessed items, this list is sorted (in-place) according to the order it came in via the original response, we do this to retain the order of the questions from the original response, as that is how they should be displayed.

### Jinja Templates and ApplicatorsResponseComponent
A new jinja template [app/assess/templates/macros/theme.jinja](https://github.com/communitiesuk/funding-service-design-assessment/blob/103a3171e3182de289ae516b16d4e2cca9033c6a/app/assess/templates/macros/theme.jinja2#L1) has been added, this template is designed to hold a mapping of all possible question/answer sub-components, and display them based on their key.  

The keys in the map correspond to concrete implementations of the abstract `ApplicatorsResponseComponent` class, new components can easily be added by creating a new concrete subclass with a unique key, said key will need added to the jinja template mapping.   

A few concrete classes have been implemented already to achieve the requirements of the [prototype](https://fsd-pre-award.herokuapp.com/s27-assessment/response-types).

Concrete classes are instantiated by a factory method on the `applicators_response.py` module, [app/assess/models/ui/applicators_response.py:117](https://github.com/communitiesuk/funding-service-design-assessment/blob/103a3171e3182de289ae516b16d4e2cca9033c6a/app/assess/models/ui/applicators_response.py#L117) - this class holds a lot of conditional logic to determine which component will be rendered for a given item, all items are fed into this factory in order and the result is what eventually goes to the front-end jinja templates.

### Test coverage

There is a lot of conditional logic in these steps, which can be difficult to maintain and modify over time - I've written tests for a variety of things like the jinja components, concrete classes & transformation steps.  I'm hoping these serve both as test coverage and a sort of documentation for future changes to this area, if we break something we should **hopefully** know by failing tests.

### Other small changes

~~Section & Criteria templates on the Assessors task list https://github.com/communitiesuk/funding-service-design-assessment/pull/58  have been modified to include the appropriate link to each sub criteria, using the first theme id as the query parameter for each given sub criteria~~

**Edit:** This has been removed in https://github.com/communitiesuk/funding-service-design-assessment/pull/73/commits/d590c7441540dff937f9ef3d4fbdaeb798d58bfa - as we're now defaulting to the first theme on the route itself, instead of passing around data and relying on hyperlinks, should also account for scenarios where user manually removes the query param, so overall better (thanks @Adam-W1)

 Corresponding assessment-store PR: https://github.com/communitiesuk/funding-service-design-assessment-store/pull/66


johnoneill46:
I've not even looked at the code yet but this is the most descriptive PR description I've ever seen - it's fantastic

app/assess/models/ui/applicators_response.py
```diff
@@ -0,0 +1,350 @@
...
+    text_items = []
+    for item in items_to_process:
+        if "answer" not in item:
+            continue
+        text_items.extend(
+            {
+                "question": remove_dashes_underscores_capitalize(answer),
+                "field_type": item.get("field_type"),
+                "field_id": item["field_id"],
+                "answer": "Yes",
```
adamdavies1, line 252: Should this be "yes"? What happens if its no?

adamdavies1:
Also the box containing applicant response should be a white background not grey on. Have a look here https://fsd-pre-award.herokuapp.com/s26-assessment/commenting/community-engagement
image

tferns:
> Also the box containing applicant response should be a white background not grey on. Have a look here https://fsd-pre-award.herokuapp.com/s26-assessment/commenting/community-engagement image

Good point.  I'll change it

app/assess/models/ui/applicators_response.py
tferns: For checkbox items, they only come through in the answers if they have been selected, so an empty list would mean they didn't select any

i.e  this has one selected checkbox, but there may have been many options

```json
{
      "answer": [
        "for-sale-or-listed-for-disposal"
      ],
      "field_id": "UDTxqC",
      "field_type": "checkboxesField",
      "form_name": "asset-information",
      "presentation_type": "list",
      "question": "Why is the asset at risk of closure?"
}
```

app/assess/models/ui/applicators_response.py
```diff
@@ -0,0 +1,398 @@
...
+        field_ids.append(tuple(field_id))
+        for fid in field_id:
+            field_ids.append(fid)
+    else:
+        field_ids.append(field_id)
+    return field_ids
+
+
+def _convert_non_number_grouped_fields(
+    response: Iterable[dict],
```
harryyo, line 313: Docstring here, what's going in and what are we expecting to get out...?

app/assess/templates/macros/theme.jinja2
```diff
@@ -0,0 +1,29 @@
...
+    'question_heading': question_heading,
+} %}
+
+{% macro theme(theme_id, answers_meta) %}
+    <h2 class="govuk-heading-m">{{ theme_id | remove_dashes_underscores_capitalize }}</h2>
+    <h3 class="govuk-heading-s response-title">Applicant's response</h3>
+    <div class="template-answer bg-white">
+        {% for meta in answers_meta %}
+            {% if meta.should_render %}
+                {{ ui_component_mapping[meta.key](meta) }}
```
harryyo, line 25: Very nice

harryyo (changes requested):
Incredible piece of work 👏 very complex functionality delivered in modular and really nicely tested code. Also great explanation in the PR notes.

Only comment for improvement, given how complex this is and fundamental it is to assessment rendering, would be to add docstrings to describe what each class/function is doing and descriptions of the expected inputs and outputs (beyond just types which you've added very nicely). Minor thing but I think worth doing for clarity also is to rename Applicators to Applicants in classes and functions (if that's what it refers to? just to keep naming similar across the repos)

But all over model work 👍

PR #76, opened 2022-12-19
FS-2024 Secure commenter routes

harryyo:
### Added global authorisation functionality via before_request function

- Ensures that all routes are protected by default via new `ensure_minimum_required_roles()` before_request function, declared in create_app factory. This ensures that all visitors have a valid signed jwt in their fsd_user_token cookie, with at least a "Commenter" role.
- Unauthenticated users are redirected from all routes to the "/" route, where they are asked to sign in
- Authenticated users with at least a "Commenter" role are redirected from "/" to "/assess/landing"
- Authenticated users with no "Commenter" role are redirected to the authenticator "roles error" message page.
- Team dashboard view has added placeholder logic to demonstrate user display logic is working as expected
- TestAuthorisation tests added

Also
- Pre-commit linting run which has created some minor layout changes in other files
- `|int|round` filter added to [application_overviews_table.html](https://github.com/communitiesuk/funding-service-design-assessment/compare/fs-2024-secure-commenter-routes?expand=1#diff-77349282b99fb3e3a91dad917986318473a6337dca09a2d84196cdee5cd0ff27) to ensure that value passed to string formatter is always a float - as this was causing a pytest error


app/create_app.py
```diff
@@ -114,6 +118,34 @@ def inject_global_constants():
         health = Healthcheck(flask_app)
         health.add_check(FlaskRunningChecker())
 
+        @flask_app.before_request
```
tferns, line 121: Beautiful!

tferns (approved):
Harry this is beautiful. couldn't spot anything to suggest!

tests/test_authorisation.py
```diff
@@ -0,0 +1,171 @@
...
+from tests.conftest import test_lead_assessor_claims
+from tests.conftest import test_roleless_user_claims
+
+
+class TestAuthorisation:
+    def test_any_unauthorised_route_redirects_to_home(self, flask_test_client):
+        """
+        GIVEN an unauthorised user
+        WHEN the user tries to access any route
+        THEN tne user is redirected to the "/" root
```
Adam-W1, line 14: lovely tests 👍  👏 
typo here on 'tne'

tests/conftest.py
```diff
@@ -54,15 +109,28 @@ def app():
...
+
+
+@pytest.fixture(scope="function")
+def flask_test_client():
+    valid_user_token = create_valid_token()
+    return next(_flask_test_client(user_token=valid_user_token))
+
+
+@pytest.fixture(scope="function")
+def flask_test_client_invalid_token():
```
Adam-W1, line 131: do we use this anywhere?

tests/test_authorisation.py
harryyo: Ah great spot! That's about a 1.5 pixel difference between n and h!

tests/conftest.py
harryyo: Ah meant to add a test for this. I've now added an additional test but removed this fixture and just updated the cookie in the test when it's called instead 👍

app/create_app.py
harryyo: Luckily someone made the before_request decorator for me...!

Adam-W1 (approved):
looks great Harry 🔥

PR #78, opened 2022-12-20
Fs 2065 Download uploaded files

adamdavies1:
This introduces a method using boto3 to get a users uploaded file for assessment by creating a pre-signed url.
Jira: https://digital.dclg.gov.uk/jira/browse/FS-2065


app/assess/models/ui/applicants_response.py
```diff
@@ -192,7 +193,8 @@ def _ui_component_from_factory(item: dict):
 
     elif presentation_type == "file":
         # TODO(FS-2065(?)): add href as link to download actual file?
-        return AboveQuestionAnswerPairHref.from_dict(item, href=answer)
+        presighned_url = get_file_url(filename=answer, application_id=application_id)
```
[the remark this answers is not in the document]
tferns: Small type on 'presighned' too.  Also if filename is None, should we be creating an AboveQuestionAnswerPair instead of an AboveQuestionAnswerPairHref?  Nothing to link to if it's None

app/assess/data.py
```diff
@@ -342,3 +346,34 @@ def get_comments(application_id: str, sub_criteria_id: str):
     )
     comment_response = get_data(comment_endpoint)
     return comment_response
+
+def get_file_url(filename: str, application_id: str):
+    """_summary_: Function is set up to retrieve
+    files from aws bucket.
+    Args:
+        filename: Takes an filename, application_id: Takes an application_id # noqa
```
harryyo, line 354: Break this onto two lines maybe? Did black wrap it? Maybe remove the comma if so?

app/assess/data.py
```diff
@@ -342,3 +346,34 @@ def get_comments(application_id: str, sub_criteria_id: str):
...
+    Returns:
+        Returns a presigned url.
+    """
+    if (Config.FLASK_ENV == "unit_test"):
+        return filename
```
tferns, line 359: Could probably mock/patch the method(s) instead of conditionals, but this is an easy option so not totally against it

harryyo (commented):
This all looks good, just a couple of typos

tferns (commented):
Looks good.  Question on how we handle scenarios where filename is None, think we should maybe not create a href element for them

app/assess/models/ui/applicants_response.py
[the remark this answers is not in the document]
tferns: Actually.  This is already handled now I re-read the code... my bad

app/assess/data.py
```diff
@@ -342,3 +346,45 @@ def get_comments(application_id: str, sub_criteria_id: str):
...
+    # Extract the JSON string from the environment variable
+        vcap_services = json.loads(os.environ["VCAP_SERVICES"])
+
+        # Check for the "aws-s3-bucket" service
+        if "aws-s3-bucket" in vcap_services:
+            # Extract the credentials from the service's configuration
+            s3_credentials = vcap_services["aws-s3-bucket"][0]["credentials"]
+            Config.AWS_ACCESS_KEY_ID = s3_credentials["aws_access_key_id"]
+            Config.AWS_SECRET_ACCESS_KEY = s3_credentials["aws_secret_access_key"]
+            Config.AWS_BUCKET_NAME = s3_credentials["bucket_name"]
```
tferns, line 373: It might be worth performing this mutation on the module itself rather than each time this function is executed, if it's scoped to the module I think it'll be executed on the first import only

tferns (approved):
Looks good! one minor suggestion

PR #86, opened 2022-12-22
Fs 2102 Comment summary

adamdavies1:
Comment summary now displaying for the correct theme and the summary page is now showing the themes the comments were made under

Also added in logic that stops the "Score the sub criteria link" being displayed on the score page

https://digital.dclg.gov.uk/jira/secure/RapidBoard.jspa?rapidView=110&projectKey=FS&view=detail&selectedIssue=FS-2102

image


app/assess/routes.py
```diff
@@ -66,9 +66,11 @@ def display_sub_criteria(
     theme_id = args.get("theme_id", sub_criteria.themes[0].id)
     fund = get_fund(Config.COF_FUND_ID)
     comments = get_comments(
-        application_id=application_id, sub_criteria_id=sub_criteria_id
+        application_id=application_id, sub_criteria_id=sub_criteria_id, theme_id=theme_id
     )
 
+    comments_dictionary = {theme.id: [comment for comment in comments if comment.theme_id == theme.id] for theme in sub_criteria.themes}
```
tferns, line 72: Little suggestion, for dictionaries I usually go with an `x_to_y` naming convention, so people typically know what's being mapped without having to read the dict comprehension, for this I'd prob go with: `theme_id_to_comments_list_map`

app/assess/templates/macros/comments_summary.html
```diff
@@ -0,0 +1,28 @@
+{% macro comment_summary(comments, themes) %}
+<div class="govuk-!-margin-top-7" id="anchor">
+    <h2 class="govuk-heading-m">Comments</h2>
+    <div id="more-detail-hint" class="govuk-hint">
+      <p class="govuk-hint">Summarise any thoughts you have on the information provided.</p>
+      <p class="govuk-hint">You cannot edit or delete your comment once you have saved it.</p>
+    </div>
+    {% if comments == None %}
```
tferns, line 8: If we changed `comments` to a `dict`, I don't think this will ever work since it will at least be an empty `dict` - maybe a truthy of `if not comments` works better?

tferns (approved):
Looks good! One minor suggestion, one thing to maybe check

Also - might be good to add some test coverage?

app/assess/templates/macros/comments_summary.html
adamdavies1: comments now returns None due to the dictionary not actually being empty even if it had no comments

PR #88, opened 2022-12-23
Fs 722 add comments

AbuNavsa:
- Can submit comments to assessment store
- Front end UI created for comments form, hidden on initial load, after clicking newly created **Add a comment** button, comment form with save button is displayed
- theme_id is added to args so the page stays on the current theme after clicking save


app/assess/data.py
```diff
@@ -377,6 +377,29 @@ def get_comments(application_id: str, sub_criteria_id: str):
...
+    }
+    url = Config.ASSESSMENT_COMMENT_ENDPOINT
+    response = requests.post(url, json=data_dict)
+    current_app.logger.info(
+        f"Response from Assessment Store: '{response.json()}'."
+    )
+    if response.status_code == 200:
+        return True
+    else:
+        return False
```
tferns, line 406: This won't work as you have the `status_code` set to `201` on your store PR: https://github.com/communitiesuk/funding-service-design-assessment-store/pull/69/files#diff-dbf73fdcc85b9741624a4d69562ff262127fe585b5b599f0e6f643f5634fc565R359-R367

Maybe could do this instead?
```suggestion
    return response.ok 
```

tferns: Although I don't think the return value is actually being used, so not a big deal

app/assess/routes.py
```diff
@@ -27,14 +28,20 @@
...
 ):
     """
     Page showing sub criteria and themes for an application
     """
+
+    displayCommentBox = False
+    if 'add-comment' in request.args.keys() and request.args['add-comment'] == '1':
+        displayCommentBox = True            
+
+    commentForm = CommentsForm()
```
tferns, line 43: Can we use `camel_case` to be consistent with the rest of the code base?

app/assess/routes.py
```diff
@@ -54,6 +61,21 @@ def display_sub_criteria(
...
+            displayCommentBox = False
+
+            submit_comment(
+                    comment=comment,
+                    application_id=application_id,
+                    sub_criteria_id=sub_criteria_id,
+                    user_id=user_id,
+                    theme_id=theme_id
+                )
+            return redirect(request.path + "?theme_id=" + theme_id)
```
tferns, line 88: Could probably use flasks built-in `url_for` for this as opposed to adding the query parameter via string concatenation

tests/test_jinja_macros.py
```diff
@@ -158,6 +159,25 @@ def test_score_macro(self, request_ctx):
...
+            commentBox=get_template_attribute("macros/comments_box.html", "commentBox"),
+            commentForm=CommentsForm()
+        )
+
+        # replacing new lines to more easily regex match the html
+        rendered_html = rendered_html.replace("\n", "")
+
+        assert re.search(
+            r"Add a comment</label>", rendered_html
+        ), "Title not found"
```
tferns, line 175: Assertion string doesn't match context

tests/test_jinja_macros.py
```diff
@@ -158,6 +159,25 @@ def test_score_macro(self, request_ctx):
...
+
+        assert re.search(
+            r"Save comment</button>", rendered_html
+        ), "Title not found"
```
tferns, line 179: Maybe should be "Save comment button not found"?

app/assess/templates/sub_criteria.html
```diff
@@ -68,6 +69,22 @@ <h2 class="govuk-heading-m">Comments</h2>
...
+                    
+                    {% if displayCommentBox != True %}
+                    <form method="GET">
+                        <button id="comment-button-community" class="govuk-button secondary-button govuk-!-margin-top-2" type="submit" data-module="govuk-button" value="add-comment">
+                            {% if comments == None %} Add a comment
+                            {% else %} Add another comment
+                            {% endif %}
+                        </button>
+                        <input type="hidden" name="theme_id" value="{{current_theme_id}}">
+                        <input type="hidden" name="add-comment" value="1">
```
tferns, line 81: Is this hidden value only used for initially hiding then showing the comment box?

If it's a boolean, we could use a checkbox instead of a pseudo bool, but I'm not too fussed tbh as it's only a minor difference

tferns (commented):
Good work on this! Think the majority of the logic is good.  I've left a couple of suggestions

app/assess/routes.py
AbuNavsa: Done and changed commentForm to comment_form as well as a few others

tests/test_jinja_macros.py
AbuNavsa: Amended

tests/test_jinja_macros.py
AbuNavsa: Amended

app/assess/data.py
AbuNavsa: Amended to response.ok

PR #98, opened 2023-01-05
FS-2119: Redirect after comment creation

tferns:
- After submitting a comment, we return a redirect to the route so it re-fetches the comments

### DIFF
--- a/app/assess/routes.py
+++ b/app/assess/routes.py
@@ -152,6 +152,15 @@ def display_sub_criteria(
                 theme_id=theme_id,
             )
 
+            return redirect(
+                url_for(
+                    "assess_bp.display_sub_criteria",
+                    application_id=application_id,
+                    sub_criteria_id=sub_criteria_id,
+                    theme_id=theme_id,
+                )
+            )
+
         return render_template(
             "sub_criteria.html",
             on_summary=False,


app/assess/routes.py
```diff
@@ -152,6 +152,15 @@ def display_sub_criteria(
...
 
+            return redirect(
+                url_for(
+                    "assess_bp.display_sub_criteria",
+                    application_id=application_id,
+                    sub_criteria_id=sub_criteria_id,
+                    theme_id=theme_id,
+                )
+            )
+
```
gio-karunakaran, line 163: 🙌🏾

gio-karunakaran (approved):
LGTM

gidsg:
@tferns this would be a good use-case for a flash message in addition to the redirect https://flask.palletsprojects.com/en/2.2.x/patterns/flashing/

tferns:
> @tferns this would be a good use-case for a flash message in addition to the redirect https://flask.palletsprojects.com/en/2.2.x/patterns/flashing/

Oh good idea! I know there was some talk of showing feedback when a user had commented so I'll see if I can fish out the story later today and whip something up

gidsg:
@tferns we probably should also redirect to a comments area anchor, else it may not be obvious to the user why they've been redirected, especially without any flash message.

tferns:
> @tferns this would be a good use-case for a flash message in addition to the redirect https://flask.palletsprojects.com/en/2.2.x/patterns/flashing/

@gidsg  I'd implemented these changes here: https://github.com/communitiesuk/funding-service-design-assessment/pull/124

Although Peter has asked design and confirmed no longer necessary, so have closed the PR

PR #102, opened 2023-01-05
FIX: Allow /healthcheck route past auth

harryyo:
### Allow /healthcheck route past auth

- Update unprotected_routes list to include /healthcheck route and remove empty route.
- Add tests to ensure homepage routes and /healthcheck route are accessible


gidsg (approved):
Nice 👍

PR #104, opened 2023-01-05
fs-2123-comments-on-score-page

gio-karunakaran:


app/assess/templates/macros/sub_criteria_navbar.html
```diff
@@ -2,7 +2,8 @@
 
 <nav class="assessment-navigation">
     <ul class="govuk-list">
-        {%for theme in sub_criteria.themes%}
+        
+        {%for theme in sub_criteria.themes|remove_theme_name_score%}
```
gidsg, line 6: This seems like a weird way to do business logic, is there a way we can pass into the template instead?

ramsharma-prog: Will add TODO to it as it may require quite few changes and few tests would fail so it was the easiest fix. But we will look into it later

PR #105, opened 2023-01-06
Adding if check to stop blank comments

adamdavies1:
You should no longer be able to add blank comments

### DIFF
--- a/app/assess/data.py
+++ b/app/assess/data.py
@@ -467,21 +467,23 @@ def get_comments(application_id: str, sub_criteria_id: str, theme_id, themes):
 def submit_comment(
     comment, application_id, sub_criteria_id, user_id, theme_id
 ):
-    data_dict = {
-        "comment": comment,
-        "user_id": user_id,
-        "application_id": application_id,
-        "sub_criteria_id": sub_criteria_id,
-        "comment_type": "COMMENT",
-        "theme_id": theme_id,
-    }
-    url = Config.ASSESSMENT_COMMENT_ENDPOINT
-    response = requests.post(url, json=data_dict)
-    current_app.logger.info(
-        f"Response from Assessment Store: '{response.json()}'."
-    )
+    # TODO: should we add something that informs the user the comment doesnt hasnt saved?
+    if comment.isspace() == False and comment != "":
+        data_dict = {
+            "comment": comment,
+            "user_id": user_id,
+            "application_id": application_id,
+            "sub_criteria_id": sub_criteria_id,
+            "comment_type": "COMMENT",
+            "theme_id": theme_id,
+        }
+        url = Config.ASSESSMENT_COMMENT_ENDPOINT
+        response = requests.post(url, json=data_dict)
+        current_app.logger.info(
+            f"Response from Assessment Store: '{response.json()}'."
+        )
 
-    return response.ok
+        return response.ok
 
 
 def get_file_url(filename: str, application_id: str):
--- a/app/assess/data.py
+++ b/app/assess/data.py
@@ -466,24 +466,22 @@ def get_comments(application_id: str, sub_criteria_id: str, theme_id, themes):
 
 def submit_comment(
     comment, application_id, sub_criteria_id, user_id, theme_id
-):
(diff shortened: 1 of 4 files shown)


stevewalton28 (approved):

app/assess/data.py
```diff
@@ -467,21 +467,23 @@ def get_comments(application_id: str, sub_criteria_id: str, theme_id, themes):
...
-        "comment_type": "COMMENT",
-        "theme_id": theme_id,
-    }
-    url = Config.ASSESSMENT_COMMENT_ENDPOINT
-    response = requests.post(url, json=data_dict)
-    current_app.logger.info(
-        f"Response from Assessment Store: '{response.json()}'."
-    )
+    # TODO: should we add something that informs the user the comment doesnt hasnt saved?
+    if comment.isspace() == False and comment != "":
```
gidsg, line 471: We should be using a form validator for this rather than by hand https://wtforms.readthedocs.io/en/2.3.x/validators/

gidsg (changes requested):
See comment, we should be using a form validator for the input, this would produce the correct flash message.

gidsg:
@adamdavies1 i.e. add a validator here https://github.com/communitiesuk/funding-service-design-assessment/blob/7186947638a792aef8e67337f47eb76af55d197a/app/assess/forms/comments_form.py#L8

app/assess/data.py
adamdavies1: Is now using validation

app/assess/forms/comments_form.py
```diff
@@ -1,8 +1,9 @@
 from flask_wtf import FlaskForm
 from wtforms import TextAreaField
 from wtforms.validators import length
+from wtforms.validators import DataRequired
 
 
 class CommentsForm(FlaskForm):
 
-    comment = TextAreaField("Comment", validators=[length(max=10000)])
+    comment = TextAreaField("Comment", validators=[length(max=5), DataRequired()])
```
gidsg: Looks like this is setting the length to max 5 characters?

app/assess/templates/macros/comments_box.html
```diff
@@ -15,6 +15,9 @@
       "classes": "govuk-label--s",
       isPageHeading: true
     },
+    "errorMessage": {
+      "text": "A comment must not be empty"
+  } if commentForm.comment.errors
```
gidsg, line 20: Could be too long or empty so this should handle both cases, can we pass through the errors we get?

app/assess/routes.py
```diff
@@ -124,7 +124,7 @@ def display_sub_criteria(
 
         comment_form = CommentsForm()
 
-        if comment_form.validate_on_submit():
+        if request.method == "POST" and comment_form.validate_on_submit():
```
harryyo, line 127: From the docs [validate_on_submit()](https://flask-wtf.readthedocs.io/en/0.15.x/quickstart/#validating-forms) also checks if the request.method is post so this shouldn't be needed I think?

app/assess/routes.py
```diff
@@ -172,7 +172,7 @@ def display_sub_criteria(
...
```
harryyo, line 175: As above

app/assess/templates/macros/comments_box.html
harryyo: Could do a catch all wording like "A comment must be between 1-1000 characters"

harryyo (changes requested):
Just a couple of comments

harryyo (approved):
👍

gidsg (approved):

app/assess/forms/comments_form.py
```diff
@@ -1,8 +1,9 @@
 from flask_wtf import FlaskForm
 from wtforms import TextAreaField
 from wtforms.validators import length
+from wtforms.validators import DataRequired
 
 
 class CommentsForm(FlaskForm):
 
-    comment = TextAreaField("Comment", validators=[length(max=10000)])
+    comment = TextAreaField("Comment", validators=[length(max=10000), DataRequired()])
```
gidsg, line 9: If you want a more friendly message, you can pass a `message` argument to the validator.

PR #108, opened 2023-01-06
fs-2128-scoring-unscored-sub-crits

gio-karunakaran:


app/assess/routes.py
```diff
@@ -144,6 +144,32 @@ def display_sub_criteria(
...
+
+            return redirect(
+                url_for(
+                    "assess_bp.display_sub_criteria",
+                    application_id=application_id,
+                    sub_criteria_id=sub_criteria_id,
+                    theme_id=theme_id,
+                )
+            )
+
```
harryyo, line 172: I think this is already handled now on line 45-65, No?

harryyo (changes requested):
One bit I think maybe can be removed...?

app/assess/routes.py
gio-karunakaran: you're right! have sorted this now 👍🏾

ramsharma-prog (changes requested):

app/assess/routes.py
```diff
@@ -144,6 +144,32 @@ def display_sub_criteria(
             (1, "Poor"),
         ]
 
+        if request.args.get("add-comment") == "1":
+            display_comment_box = True
```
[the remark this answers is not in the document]
gio-karunakaran: sorted!

harryyo (approved):
👍

ramsharma-prog (approved):
LGTM 😄

PR #110, opened 2023-01-06
BAU: Show application status as Flagged when flag exists

tferns:
- Tiny change to override status as "Flagged" when an application has a flag

image


harryyo (approved):
👍

PR #112, opened 2023-01-09
Removing placeholder metrics for lead assessors for launch

stevewalton28:
To be addressed as part of FS-1666 - no need for placeholder data in the markup


gidsg (approved):
If this markup is useful should we comment it out rather than deleting it? e.g. https://jinja.palletsprojects.com/en/3.1.x/templates/#comments

PR #115, opened 2023-01-09
Fix for file downloads

adamdavies1:
Work around for the pre signed url not downloading the file on uat

### DIFF
--- a/app/assess/data.py
+++ b/app/assess/data.py
@@ -507,11 +507,14 @@ def get_file_response(file_name: str, application_id: str):
 
     try:
         obj = s3_client.get_object(Bucket=Config.AWS_BUCKET_NAME, Key=prefixed_file_name)
-
+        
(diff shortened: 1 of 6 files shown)


app/assess/data.py
```diff
@@ -482,38 +483,36 @@ def submit_comment(
...
-                "Key": prefixed_file_name,
-            },
-            ExpiresIn=3600,
-        )
+        obj = s3_client.get_object(Bucket=Config.AWS_BUCKET_NAME, Key=prefixed_file_name)
 
+        data = obj['Body'].read()
+
+    # Create a Flask response object
+        response = Response(data, mimetype='text/plain', headers={'Content-Disposition': f'attachment;filename={file_name}'})
```
gidsg, line 514: Don't think we should be setting the mimetype to 'text/plain` as response won't be plain text necessarily. Can we try removing the mimetype, or using the content type we  get back from AWS?

gidsg (approved):

PR #119, opened 2023-01-10
FS-2136: Set sub criteria width to one half

tferns:
- Set sub criteria width to one half, should enable consistency even with long names

**Before:**

image

**After:**

image


tferns:
Note:  this will also collapse onto a new line if the title is longer

image

harryyo (approved):

PR #126, opened 2023-01-11
Banner completed

AbuNavsa:
Banner for STOPPED Assessment Completed

-  Assessment Stopped Header displayed in place of "Flagged" when an application has a "STOPPED"  flag
- The banner summary also now has "Stopped" in place of the workflow_status if the flag is present
- Tests have been written to check if the UI is showing accurate to Flagged/Assessment Stopped heading
- Banner_summary now has a default parameter of flag=None, and the banner only adjust to the stopped flag is this flag is not None


app/assess/templates/assessor_tasklist.html
```diff
@@ -17,15 +17,15 @@
     </div>
 
     {% if flag and g.user.highest_role in ("ASSESSOR", "LEAD_ASSESSOR") %}
-        {{ assessment_flag(flag, flag_user_info) }}
+        {{ assessment_flag(flag) }}
```
tferns, line 20: Don't we need the user info here?

app/assess/routes.py
```diff
@@ -289,7 +290,7 @@ def application(application_id):
         state=state,
         application_id=application_id,
         flag=flag,
-        flag_user_info=accounts.get(flag.user_id) if flag else None,
+        # flag_user_info=accounts.get(flag.user_id) if flag else None,
```
tferns, line 293: Should this be commented out?

app/assess/templates/macros/banner_summary.html
```diff
@@ -13,7 +13,15 @@ <h3 class="govuk-body-l fsd-banner-content fsd-banner-collapse-padding">Project
                         name: {{ project_name }}</h3>
                     <h3 class="govuk-body-l fsd-banner-content">Total funding requested:
                         &pound;{{ "{:,.2f}".format(funding_amount_requested | float) }}</h3>
-                    <h3 class="govuk-body-l fsd-banner-content fsd-banner-collapse-padding">{{ workflow_status | status_to_human }}</h3>
+                    <h3 class="govuk-body-l fsd-banner-content fsd-banner-collapse-padding">
+                        {% if flag == None  %}
+                            {{ workflow_status | status_to_human }}
+                        {% else %}
+                            {{ flag.flag_type.name | capitalize }}
+                        {% endif %}
```
tferns, line 21: I'd probably flip this and go with:
```suggestion
                        {% if flag %}
                            {{ flag.flag_type.name | capitalize }}
                        {% else %}
                             {{ workflow_status | status_to_human }}
                        {% endif %}
```

Seems more intuitive to me, but to each their own (not a big deal)

app/assess/templates/macros/banner_summary.html
```diff
@@ -1,4 +1,4 @@
-{% macro banner_summary(fund_name, project_reference, project_name, funding_amount_requested, workflow_status) %}
+{% macro banner_summary(fund_name, project_reference, project_name, funding_amount_requested, workflow_status, flag=None) %}
```
tferns, line 1: Did you also add a change somewhere that calls `banner_summary` with a flag? I can't see it in this PR

tferns (approved):
Changes look good but think you're missing passing flag to `banner_summary`

tferns (changes requested):
Actually sorry to revoke the approve but I think we need `user_info` re-added too

app/assess/templates/assessor_tasklist.html
AbuNavsa: Yup, removed temporarily to test, thought I put it back!

app/assess/routes.py
AbuNavsa: Will be removed

app/assess/templates/macros/banner_summary.html
AbuNavsa: You're right!

AbuNavsa:
Done in new branch

PR #127, opened 2023-01-11
BAU: Only call progress endpoint if we have data

tferns:
- Only call the progress endpoint if we actually have data, if we don't, no point in calling it

Fixes: 
- https://communities-govuk.slack.com/archives/C02PHBER287/p1673449332218439
- https://sentry.io/organizations/funding-service-design-team-dl/issues/3850987122/?referrer=slack

### DIFF
--- a/app/assess/routes.py
+++ b/app/assess/routes.py
@@ -233,12 +233,11 @@ def landing():
     assessment_deadline = get_round(
         Config.COF_FUND_ID, Config.COF_ROUND2_ID
     ).assessment_deadline
-    
+
     # Updating assessment progress for applications
-    application_overviews = get_assessment_progress(
-        application_overviews
-        )
-  
+    if application_overviews:
+        application_overviews = get_assessment_progress(application_overviews)
+
     return render_template(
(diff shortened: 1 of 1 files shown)


app/assess/routes.py
```diff
@@ -233,12 +233,11 @@ def landing():
...
     ).assessment_deadline
-    
+
     # Updating assessment progress for applications
-    application_overviews = get_assessment_progress(
-        application_overviews
-        )
-  
+    if application_overviews:
+        application_overviews = get_assessment_progress(application_overviews)
```
gidsg, line 239: Bit confusing as looks like we're mutating this? Can we call it something different rather than mutate?

tferns: Yeah good point.

tferns: Addressed in https://github.com/communitiesuk/funding-service-design-assessment/pull/127/commits/d45232f3ee9c362c215d467060038b4ac9d9fb0d

gidsg (approved):

PR #128, opened 2023-01-11
FS-2092-mark-assessment-complete

ramsharma-prog:
Once all the sub_criterias are scored, a post request is made to the assessment-store to update the assessment status for that application to "COMPLETED" and the status on the assessment dashboard is also updated to indicate that it has been completed.

PR for assessment-store:
https://github.com/communitiesuk/funding-service-design-assessment-store/pull/101


app/assess/display_value_mappings.py
```diff
@@ -22,4 +22,5 @@
     "QA_COMPLETE": "QA Complete",
     "FLAGGED": "Flagged",
     "INELIGIBLE": "Ineligible",
+    "COMPLETED": "Completed",
```
tferns, line 25: This should update the filter on the front-end Team dashboard to include "Completed"

This isn't in the prototype: https://fsd-pre-award.herokuapp.com/s26-assessment/dashboard

Although I don't see any issue with having it... will depend on design.  Also, if we just mark as completed, when does the QA related things take place? Is that done by different work in flags?

tferns: If we end up leaving this in the front-end, can we make sure the search filter functionality actually works with the "Completed" status?  

I think it will judging by how the logic is in the assessment-store: https://github.com/communitiesuk/funding-service-design-assessment-store/blob/909ae0796a73ada165a23a1ea401b0cb2e3cb323/db/queries/assessment_records/queries.py#L60-L64

But worth checking in-case.  If you find a way to remove it from the UI to match the prototype, then we don't need to worry about it though.

app/assess/routes.py
```diff
@@ -289,16 +293,27 @@ def application(application_id):
...
         accounts = get_bulk_accounts_dict([flag.user_id])
-
+        
+    sub_criteria_status_completed = all_status_completed(state)
+    form = AssessmentCompleteForm()
+    if request.method=="POST":
+        requests.post(
+            Config.ASSESSMENT_UPDATE_STATUS.format(
+                application_id=application_id)
+            )
```
tferns, line 303: Indentation is messed up here, could you install the pre-commit hooks so they run when making changes?




From the root of the project run:
```
pre-commit install 
```

Otherwise other people commits will fix the formatting and it will pollute their PRs with unrelated changes, this should also help with merge conflicts we'll be using the same format & have sorted imports, so less work for others when solving conflicts.

app/assess/routes.py
```diff
@@ -289,16 +293,27 @@ def application(application_id):
...
         application_id=application_id,
         flag=flag,
+        show_prompt= False,
         flag_user_info=accounts.get(flag.user_id) if flag else None,
-    )
-
-
+)
+        
+
```
tferns, line 316: Trailing whitespace

app/assess/routes.py
```diff
@@ -289,16 +293,27 @@ def application(application_id):
...
+            )
+   
     return render_template(
         "assessor_tasklist.html",
+        sub_criteria_status_completed = sub_criteria_status_completed,
+        form = form,
         state=state,
         application_id=application_id,
         flag=flag,
+        show_prompt= False,
```
tferns, line 312: Is this needed if it's always going to be false?

tferns (commented):
I think the logic in this pull request looks good, I've left some comments or consideration though

app/assess/routes.py
ramsharma-prog: I am alreday tiding up as we speak 😄

app/assess/routes.py
ramsharma-prog: Hmm, i think i may have added it when i was trying to use JS but forgot to remove. Wil remove it

app/assess/routes.py
ramsharma-prog: I am rinning pre commit on whole file once I tided up

app/assess/routes.py
tferns: 👍 IF you install the hooks this gets done for you when committing

gio-karunakaran (approved):
LGTM

PR #130, opened 2023-01-11
FS-2155, FS-2157: Fix score styling to match proto

tferns:
- Formatted the highest_role using `all_caps_as_human`
- Add margins to match prototype spacing
- Change button names to match prototype
- Add classes to change button styling to prototype
- Add styling for latest score box

Video demo:
https://user-images.githubusercontent.com/117724519/211872026-01234017-f598-40cb-b84f-0974d27f4164.mp4


harryyo (approved):
👍

PR #131, opened 2023-01-12
add logic to resolve a flag

Adam-W1:
https://digital.dclg.gov.uk/jira/browse/FS-2018
Resolving a flagged application 

### DEPENDS ON:
- [Assessment store PR](https://github.com/communitiesuk/funding-service-design-assessment-store/pull/98)

**Resolve on flag application (for lead assessor only)**

Pasted Graphic

**Two required options to resolve a flag with error handling** 

Pasted Graphic 3

**When resolved, a user can re-flag**

Pasted Graphic 4

**User can stop and see this flag and dashboard**

Pasted Graphic 6

**Field required validation**
image


app/assess/models/flag.py
```diff
@@ -8,18 +8,14 @@ class FlagType(Enum):
     FLAGGED = 0
     STOPPED = 1
     QA_COMPLETED = 2
-
-
-class ResolutionType(Enum):
-    QUERY_RESOLVED = 0
-    STOP_ASSESSMENT = 1
+    RESOLVED = 3
```
gidsg, line 11: Does this need a corresponding database migration?

app/assess/models/flag.py
```diff
@@ -8,18 +8,14 @@ class FlagType(Enum):
...
 
 
 @dataclass()
 class Flag:
+    id: str
```
gidsg, line 16: could this be a UUID type?

app/assess/models/flag.py
Adam-W1: yes it's here: https://github.com/communitiesuk/funding-service-design-assessment-store/pull/98/files#diff-ec0465a8bc7a7def015e2569574477919c9d91d1971736f1fc233249e023f22c

gidsg: As that is not merged yet, shall we make this draft until that is merged, because this won't work until then?

Adam-W1: yes sure 👍

app/assess/models/flag.py
Adam-W1: So this is a pattern across all the dataclasses on the frontend. Its a uuid on the [datamodel class](https://github.com/communitiesuk/funding-service-design-assessment-store/pull/98/files#:~:text=%3D%20%22flags%22-,id%20%3D%20db.Column(,),-justification%20%3D%20db) in the database but on the frontend we just pass it around as a string until it returns. back to the store (we dont actually use the flag id yet anyway)

app/assess/forms/resolve_flag_form.py
```diff
@@ -0,0 +1,14 @@
...
+
+class ResolveFlagForm(FlaskForm):
+    resolution_flag = RadioField(
+        "resolve", 
+        choices=["RESOLVED", "STOPPED"], 
+        validators=[DataRequired()]
+    )
+    justification = TextAreaField(
+        "justification",
+        validators=[DataRequired()],
```
gidsg, line 13: For text area can we use `InputRequired`?

Adam-W1: good spot 👍 
context: Certain field types coerce data into Falsey values(0, Decimal(0), etc.). The problem occurs when you have an IntegerField and the form submits a value like '0'. If you apply DataRequired to this it will fail validation. This is because DataRequired will evaluate if not field.data... after coercion where field.data is the Falsey numeric value 0.

Adam-W1: With this in mind, I could change the other data_required validators to input_required as part of this PR?

gidsg: Yes, maybe I think the recommendation is it to use this instead of `DataRequired`

tests/test_routes.py
```diff
@@ -352,19 +400,41 @@ def test_flag_route_submit_flag(
...
+    def test_flag_route_resolve_flag(
+        self,
+        flask_test_client,
+    ):
+        token = create_valid_token(test_lead_assessor_claims)
+        flask_test_client.set_cookie("localhost", "fsd_user_token", token)
+
+        response = flask_test_client.get(
+            "assess/resolve_flag/app_123?section=org_info",
+        )
```
gidsg, line 434: Are we testing the POST to this route?

app/assess/forms/resolve_flag_form.py
Adam-W1: cool - ive rolled out to other forms 👍

tests/test_routes.py
Adam-W1: Added a test for a POST too :]

gidsg (approved):

PR #132, opened 2023-01-12
BAU: Change progress endpoint to post

tferns:
- Start posting data as a json body as opposed to query params

Fixes: https://sentry.io/organizations/funding-service-design-team-dl/issues/3850987122/?referrer=slack

### DIFF
--- a/app/assess/data.py
+++ b/app/assess/data.py
@@ -56,14 +56,23 @@ def get_assessment_progress(application_metadata):
         ]
     }
     endpoint_url = Config.ASSESSMENT_PROGRESS_ENDPOINT
-    response = get_data(endpoint=endpoint_url, payload=application_ids_list)
-    if response is not None:
-        [
-            x.update({"progress": res.get("progress")})
-            for res in response
-            for x in application_metadata
-            if res["application_id"] == x["application_id"]
-        ]
+    current_app.logger.info(
+        f"Fetching assessment progress from '{endpoint_url}', with json"
+        f" payload: {application_ids_list}."
+    )
+    response = requests.post(endpoint_url, json=application_ids_list)
+
+    if not response.ok:
+        current_app.logger.error(
+            f"Could not get assessment progress for endpoint '{endpoint_url}'"
+        )
+        return application_metadata
+
+    response_json = response.json()
+    for res in response_json:
+        for am in application_metadata:
+            if am.get("application_id") == res.get("application_id"):
+                am["assessment_progress"] = res.get("assessment_progress")
 
     return application_metadata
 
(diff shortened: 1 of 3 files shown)


app/assess/routes.py
```diff
@@ -235,9 +235,13 @@ def landing():
...
     post_processed_overviews = (
-        get_assessment_progress(application_overviews)
-        if application_overviews
-        else []
+        (
+            get_assessment_progress(application_overviews)
+            if application_overviews
+            else []
+        )
+        if not Config.USE_LOCAL_DATA
```
tferns, line 243: This is a bit hacky.  It's way for the tests to not call the post request since we don't have any `get_local_data` concept for post requests setup at the moment.

tests/test_initialisation.py
```diff
@@ -16,6 +16,11 @@ def test_flask_initiates(flask_test_client):
     If this test succeeds then our flask application
     is AT LEAST up and running without errors.
     """
+    flask_test_client.set_cookie(
+        "localhost",
+        "fsd_user_token",
+        "",
+    )
```
tferns, line 23: This was added as we don't want the default cookie used here, added at request of @harryyo

app/assess/data.py
```diff
@@ -525,16 +534,21 @@ def get_file_response(file_name: str, application_id: str):
...
+            headers={
+                "Content-Disposition": f"attachment;filename={file_name}"
+            },
+        )
         return response
     except ClientError as e:
         current_app.logger.error(e)
-        raise Exception(e)
\ No newline at end of file
+        raise Exception(e)
```
tferns: Pre-commit formatting

gidsg (approved):
👍

app/assess/routes.py
gidsg: Can we maybe add a TODO/ticket for this so it doesn't get forgotten?

harryyo: Yeah hmmm it does feel a bit dirty having this in the main code... Maybe as it is a temporary fix to enable tests to pass perhaps the workaround should go in the tests file instead (with a TODO to note it). Also have you added a ticket to do the full fix (if not time to do it now)?

harryyo (commented):
Just a note about adding a TODO and tech-debt ticket for post

app/assess/routes.py
tferns: Added in https://github.com/communitiesuk/funding-service-design-assessment/pull/132/commits/bca54c2337d16c746cb7d2318dcfbfc54ff5458b, thanks

tferns:
> Just a note about adding a TODO and tech-debt ticket for post

FYI @harryyo created https://digital.dclg.gov.uk/jira/browse/FS-2161 for addressing this

PR #135, opened 2023-01-12
FS-2111: Update application overview table to display Country in locations column

johnoneill46:
https://digital.dclg.gov.uk/jira/browse/FS-2111

As part of FS-2111 we require the Country to be displayed on the Assessment FE

(Any data that is retrieved as null will be displayed as an empty string on the FE - see screenshot attached for example) 

Screenshot 2023-01-12 at 16 01 55


app/assess/templates/components/application_overviews_table.html
```diff
@@ -102,6 +105,7 @@
                 <td class="govuk-table__cell"><a class="govuk-link" href="{{url_for('assess_bp.application',application_id=overview.application_id)}}">{{overview.project_name}}</a></td>
                 <td class="govuk-table__cell">{{asset_types[overview.asset_type]}}</td>
                 <td class="govuk-table__cell">&pound;{{ "{:,.2f}".format(overview.funding_amount_requested|int|round) }}</td>
+                <td class="govuk-table__cell">{{ overview.location_json_blob.country }}</td>
```
tferns, line 108:
```suggestion
                <td class="govuk-table__cell">{{ overview.location_json_blob.country if overview.location_json_blob else 'N/A'}}</td>
```

Could change `N/A` to a blank string or something else

johnoneill46: Change made as per suggestion 👍

app/assess/templates/components/application_overviews_table.html
```diff
@@ -102,6 +105,7 @@
                 <td class="govuk-table__cell"><a class="govuk-link" href="{{url_for('assess_bp.application',application_id=overview.application_id)}}">{{overview.project_name}}</a></td>
                 <td class="govuk-table__cell">{{asset_types[overview.asset_type]}}</td>
                 <td class="govuk-table__cell">&pound;{{ "{:,.2f}".format(overview.funding_amount_requested|int|round) }}</td>
+                <td class="govuk-table__cell">{{ overview.location_json_blob.country if overview.location_json_blob else 'N/A'}}</td>
```
gidsg, line 108: Can we make this just empty string or remove the `else`?

gidsg:
Being a bit nit-picky here but can you remove `as json` from the PR name, as we're not displaying them as json if I understand correctly.

stevewalton28 (dismissed):

tests/api_data/endpoint_data.json
```diff
@@ -357,7 +357,7 @@
       "type_of_application": "sporting",
       "project_name": "Save the rugby pitch",
       "funding_amount_requested": "25000",
-      "location": "Scotland"
+      "location_json_blob": "Scotland"
```
gidsg, line 360: Shouldn't this be JSON data rather than a string?

johnoneill46:
> Being a bit nit-picky here but can you remove `as json` from the PR name, as we're not displaying them as json if I understand correctly.

Totally agree - Done

tests/api_data/endpoint_data.json
johnoneill46: Good spot - fixed accordingly

app/assess/templates/components/application_overviews_table.html
```diff
@@ -105,7 +105,7 @@
                 <td class="govuk-table__cell"><a class="govuk-link" href="{{url_for('assess_bp.application',application_id=overview.application_id)}}">{{overview.project_name}}</a></td>
                 <td class="govuk-table__cell">{{asset_types[overview.asset_type]}}</td>
                 <td class="govuk-table__cell">&pound;{{ "{:,.2f}".format(overview.funding_amount_requested|int|round) }}</td>
-                <td class="govuk-table__cell">{{ overview.location_json_blob.country if overview.location_json_blob else 'N/A'}}</td>
+                <td class="govuk-table__cell">{{ overview.location_json_blob.country}}</td>
```
gidsg, line 108: I still think we need the `if` statement here, as won't we get an exception if the data is not populated yet?

tests/api_data/endpoint_data.json
```diff
@@ -471,7 +513,13 @@
       "type_of_application": "sporting",
       "project_name": "Save the rugby pitch",
       "funding_amount_requested": "25000",
-      "location_json_blob": "Scotland"
+      "location_json_blob": {
+        "error": false,
```
gidsg, line 517: what is this error key for?

johnoneill46: It's what is returned from postcodes.io to let you know it's a valid postcode - means if needed in the future we can quickly filter out or identify invalid postcodes

app/assess/templates/components/application_overviews_table.html
johnoneill46: No it just displays as blank (see the screenshot in the PR description above ^)

gidsg:
Please can we add a test for the view for both the data being populated and not populated?

app/assess/templates/components/application_overviews_table.html
gidsg: Would have more confidence in this if we had a test and the test data included some where the field is `None`, as per https://github.com/communitiesuk/funding-service-design-assessment/pull/135#issuecomment-1381723634

johnoneill46: Tests to be added to the backend as part of the mapper that gets the location data:
https://github.com/communitiesuk/funding-service-design-assessment-store/pull/107/files

gidsg (approved):
LGTM 👍 (as we don't have test coverage on the values being none, we're trusting flask can handle this)

PR #138, opened 2023-01-16
FS-2166 Fixing axe error for justification 

adamdavies1:
Fixing the axe error for no label on justification

### DIFF
--- a/app/assess/templates/macros/justification.html
+++ b/app/assess/templates/macros/justification.html
@@ -4,7 +4,7 @@
 
 <div class="govuk-form-group comments {% if justification_error %}govuk-form-group--error{% endif %}" name="just-text-area" id="just-text-area">
     <h2 class="govuk-label-wrapper">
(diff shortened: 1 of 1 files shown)


app/assess/templates/macros/justification.html
```diff
@@ -17,7 +17,7 @@ <h2 class="govuk-label-wrapper">
       </p>
     {% endif %}
     <textarea class="govuk-textarea" id="justification" name={{justification_form_name}} value=""
-        rows="10" aria-describedby="more-detail-hint"></textarea>
+        rows="10" aria-describedby="more-detail-hint" aria-label="Justification"></textarea>
```
gidsg, line 20: Can we use `label for` instead and link it to the label instead? e.g. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

adamdavies1: Yes we can. Have pushed this change

gidsg (approved):

PR #139, opened 2023-01-17
Fs 2008 stopped banner

AbuNavsa:
Banner for STOPPED Assessment Completed

- Assessment Stopped Header displayed in place of "Flagged" when an application has a "STOPPED" flag
- The banner summary also now has "Stopped" in place of the workflow_status if the flag is present
- Tests have been written to check if the UI is showing accurate to Flagged/Assessment Stopped heading
- Banner_summary now has a default parameter of flag=None, and the banner only adjust to the stopped flag is this flag is not None

### DIFF
--- a/app/assess/data.py
+++ b/app/assess/data.py
@@ -368,7 +368,7 @@ def get_latest_flag(application_id: str) -> list[Flag] | None:
         )
     )
     if flag:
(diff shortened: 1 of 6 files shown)


gio-karunakaran (approved):
LGTM 👍🏾

PR #140, opened 2023-01-17
Fix for commas in file names

adamdavies1:
Fixing bug that stopped downloading of files when the name had commas in them.


gidsg (approved):
LGTM 👍 (would be nice to unit test this if possible)

PR #141, opened 2023-01-17
Restart stopped application added with tests (auth added)

Adam-W1:
As a Lead Assessor 

I want to be able to restart a Stopped application

So that an application can be assessed when it is eligible / this is possible

https://digital.dclg.gov.uk/jira/browse/FS-2093


config/envs/development.py
```diff
@@ -19,9 +20,13 @@ class DevelopmentConfig(DefaultConfig):
...
     # for local testing with flask run and USE_LOCAL_DATA = True:
     # USE_LOCAL_DATA = True
+    # if USE_LOCAL_DATA:
+    #     ACCOUNT_STORE_API_HOST = CommonConfig.TEST_ACCOUNT_STORE_API_HOST
+    #     APPLICATION_STORE_API_HOST = CommonConfig.TEST_APPLICATION_STORE_API_HOST
+    #     ASSESSMENT_STORE_API_HOST = CommonConfig.TEST_ASSESSMENT_STORE_API_HOST
+    #     FUND_STORE_API_HOST = CommonConfig.TEST_FUND_STORE_API_HOST
 
     # for testing with docker runner:
-    USE_LOCAL_DATA = False
```
gidsg, line 24: did we mean to include these config changes or were they just for local testing?

Adam-W1: yes I popped them in on purpose, we need a ref to the local data when running locally (without docker) so all that devs need to do is uncomment this code block, and it will call test data.

harryyo: Doesn't CommonConfig already default to these if not env var set? eg. https://github.com/communitiesuk/funding-service-design-utils/blob/main/fsd_utils/config/commonconfig.py#L42

Adam-W1: It does, but in docker runner if you want to run off local test data we need to restate it otherwise it keeps trying to call the other containers which are picked up in common config (https://github.com/communitiesuk/funding-service-design-docker-runner/blob/main/docker-compose.yml#:~:text=fund%2Dstore%3A8080-,%2D%20APPLICATION_STORE_API_HOST%3Dhttp%3A//application%2Dstore%3A8080,%2D%20ACCOUNT_STORE_API_HOST%3Dhttp%3A//account%2Dstore%3A8080,-ports%3A)

app/assess/templates/macros/assessment_flag.html
```diff
@@ -1,20 +1,29 @@
...
+        <h1 class="assessment-alert__heading govuk-heading-l">Assessment {{ flag.flag_type.name | title }}</h1>
+        <h2 class="govuk-heading-m">Reason</h2>
+        <p class="govuk-body">{{ flag.justification }}</p>
+        <p class="govuk-body-s">{{ user_info["full_name"] }} ({{ user_info["highest_role"] | all_caps_to_human }}) {{ user_info["email_address"] }}</p>
+        <p class="govuk-body-s">{{ flag.date_created | datetime_format("%d/%m/%Y at %H:%M") }}</p>
+        {% if current_user_role == "LEAD_ASSESSOR"  %}
+            <p class="govuk-body"><a class="govuk-link govuk-link--no-visited-state" href="{{url_for('assess_bp.continue_assessment',application_id=application_id)}}">Remove flag and continue assessment</a></p>
+        {% endif %}
+    </div>
+{% endmacro %}
```
tferns, line 29: Quite a lot of duplication here.  But I think there is enough differences to justify the way it's been done, so I'm not too fussed on this personally, if there were 4 or more types then I'd think using parameters would be better... but that might be something to do in the future rather than now.

tests/test_authorisation.py
```diff
@@ -351,9 +340,136 @@ def test_different_user_levels_have_correct_flagging_permissions(
                     "assess/flag/app123",
                     follow_redirects=True,
                 )
+                assert False, "Did not throw expected error."
```
tferns, line 343: If you're interested, there is a better way to test expected exceptions using a built-in pytest context manager:

https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions

Here's an example I used this for before:

https://github.com/communitiesuk/funding-service-design-assessment/blob/02bc025a91a15e79c9d1d21951f08c298601f67a/tests/test_applicators_response.py#L373-L382

config/envs/development.py
Adam-W1: using the above you can turn on and off what you want to be mocked and what you dont

tests/test_authorisation.py
```diff
@@ -351,9 +340,136 @@ def test_different_user_levels_have_correct_flagging_permissions(
...
                 )
+
+    @pytest.mark.parametrize(
+        "claim,expect_resolve_flag_available",
+        [
+            (test_commenter_claims, False),
+            (test_assessor_claims, False),
+            (test_lead_assessor_claims, True),
+        ],
+    )
```
tferns, line 358: There is a boolean trap here, with some logic built-in to the test

My preference is usually to split these into two separate tests, rather than maintain the logic

I'm not too fussed about this, but let me know what you think

tests/test_authorisation.py
```diff
@@ -351,9 +340,136 @@ def test_different_user_levels_have_correct_flagging_permissions(
...
+                    "assess/resolve_flag/app123",
+                    follow_redirects=True,
+                )
+                assert False, "Did not throw expected error."
+            except Exception as e:
+                # Redirect to authenticator
+                assert (
+                    e.args[0]
+                    == "Following external redirects is not supported."
+                )
```
tferns, line 402: If you end up keeping this logic, and not splitting them into two separate tests, but you want to use pytest.rasises, - you can parametrize the context manager and use contextlib.nullcontext() for exception scenarios (let me know if you want me to elaborate on this, or give some examples)

tferns (approved):
Really nice work on this.  I think this work is really solid and well done, but I've left some comments of some opinionated quality of life/best practice improvements, if you're interested in incorporating them.  Also happy to discuss/elaborate on them if you're curious.

app/assess/forms/continue_application_form.py
```diff
@@ -0,0 +1,10 @@
+from flask_wtf import FlaskForm
+from wtforms import TextAreaField
+from wtforms.validators import InputRequired
+
+
+class ContinueApplicationForm(FlaskForm):
+    reason = TextAreaField(
+        "reason",
+        validators=[InputRequired()],
+    )
```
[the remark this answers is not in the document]
Adam-W1: good shout 👍 adding

app/assess/templates/macros/assessment_flag.html
tferns: Yeah I think the [rule of 3](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) is a good baseline to go from, I've spent a lot of time in the past unwinding pre-mature abstractions so I'm quite lax on duplication until there's an obscene amount of it

tests/test_authorisation.py
Adam-W1: Im not too concerned about it as it's test code (i agree diff story in prod code), the reason I have parametrised over writing individual tests it to keep the line count down and keep the test files readable and extendable if we want to test for more roles in the future (or switch the permissions of certain roles). I think it would be easier to go in and switch the bool rather than rewrite each test.

tests/test_authorisation.py
Adam-W1: ill give your suggestion a go, looks cleaner and more intuitive 👍

tests/test_authorisation.py
tferns: Yeah reading the full test in isolation I actually like the way you've done it.  I think if we had separate tests we would still parametrize the claims, but the boolean logic would just be two separate tests so it would be shuffling from one tests parametrized claims to anothers - but let's leave it how it is cause it's not a big deal and looks good as it is just now.

PR #143, opened 2023-01-18
FS-2210-confirmation-banner-and-status-to-match-prototype

ramsharma-prog:
- The assessment status is labeled "Assessment Complete" on the tasklist banner if the assessment has been completed.
- CSS has been added for the "Assessment Complete" prompt.

Screenshot 2023-01-18 at 11 09 27

### DIFF
--- a/app/assess/templates/assessor_tasklist.html
+++ b/app/assess/templates/assessor_tasklist.html
@@ -53,13 +53,12 @@
     {% if flag and g.user.highest_role in ("ASSESSOR", "LEAD_ASSESSOR") and not state.flag_resolved %}
         {{ assessment_flag(flag, flag_user_info, application_id) }}
     {% endif %}
-
     {{ banner_summary(
         state.fund_name,
         state.short_id,
         state.project_name,
         state.funding_amount_requested,
-        state.workflow_status,
+        'ASSESSMENT_COMPLETE' if state.workflow_status=='COMPLETED' else state.workflow_status,
         flag
     ) }}
 
--- a/app/assess/templates/assessor_tasklist.html
+++ b/app/assess/templates/assessor_tasklist.html
@@ -58,7 +58,7 @@
         state.short_id,
         state.project_name,
         state.funding_amount_requested,
-        'ASSESSMENT_COMPLETE' if state.workflow_status=='COMPLETED' else state.workflow_status,
+        state.workflow_status,
         flag
     ) }}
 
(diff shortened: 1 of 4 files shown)


app/assess/templates/assessor_tasklist.html
```diff
@@ -53,13 +53,12 @@ <h2 class="assessment-alert__heading govuk-heading-l govuk-!-margin-top-4">All s
...
         {{ assessment_flag(flag, flag_user_info, application_id) }}
     {% endif %}
-
     {{ banner_summary(
         state.fund_name,
         state.short_id,
         state.project_name,
         state.funding_amount_requested,
-        state.workflow_status,
+        'ASSESSMENT_COMPLETE' if state.workflow_status=='COMPLETED' else state.workflow_status,
```
tferns, line 61: Can't we just use COMPLETED? Is there a need for this logic?

app/assess/views/filters.py
```diff
@@ -8,6 +8,7 @@ def status_to_human(status: str):
         "SUBMITTED": "Submitted",
         "COMPLETED": "Completed",
         "FLAGGED": "Flagged",
+        "ASSESSMENT_COMPLETE": "Assessment complete"
```
tferns, line 11: Can't we just use COMPLETED? Is there a need for this logic?

app/assess/templates/assessor_tasklist.html
ramsharma-prog: Bug was raised to change it to 'Assessment complete'. Although I had it as 'Completed' originally

app/assess/views/filters.py
ramsharma-prog: Bug was raised to change it to 'Assessment complete'. Although I had it as 'Completed' originally

tferns: But can't we keep the key the same and just change the value? As in change L9 to "Assessment complete"

ramsharma-prog: Wasn't sure if the status "Completed was being used somewhere else so added new line ?

tferns: Yeah good point I think its being used elsewhere on criterias like you said:
https://github.com/communitiesuk/funding-service-design-assessment/blob/c783013bd61b086b39d721e1e800985d00b09380/app/assess/templates/macros/criteria_element.html#L18

Should we move the logic inside the banner_summary macro so it applies to all pages instead of the task list?  
Can we also check that if the status is completed the banner state is shared across subpages i.e sub criteria?

ramsharma-prog: Yes, we could move the logic to banner_summary. 
Here is the prototype: https://fsd-pre-award.herokuapp.com/s26-assessment/assessment-overview. I think the value should be "Complete" for "COMPLETED" instead of "Completed"

gio-karunakaran (approved):
LGTM

PR #146, opened 2023-01-19
BAU: fix score label

gidsg:
Label for attribute needs to match input ID to be usable. This meant text label isn't clickable and wouldn't have been associated by screenreaders. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label for context on HTML.

## Before
https://user-images.githubusercontent.com/1764158/213491551-99365b25-21a0-4a2f-8b60-814372efc600.mov

## After
https://user-images.githubusercontent.com/1764158/213491609-655bcad4-38d6-4441-8d26-b0242ea48528.mov


gio-karunakaran (approved):
LGTM

PR #147, opened 2023-01-20
Stopped supercedes flagged

AbuNavsa:
Stopped flag shows as status if it is the most recent flag. After resolved, it will show the workflow_status.


Adam-W1 (changes requested):
Looks good Abu (and thanks for adding the extra changes) 👍 I think Its worth adding a test (adding sample test data to test.json) and making sure we load the statuses correctly on the dashboard

Adam-W1 (approved):
good work Abu 👍

PR #150, opened 2023-01-23
FS-2177-save-score-rationale-without-scores

ramsharma-prog:
The "justification" macro is designed to save any rationale provided by the user, even if the assessment has not been scored yet. This means that if the user enters text into the text area for the rationale, but does not provide a score for the assessment, the page will still save the rationale text entered by the user.


rayhan15243 (approved):

gio-karunakaran (approved):
LGTM

srh-sloan (approved):
Good to go 👍

PR #151, opened 2023-01-24
separate helpers and filter tests, add dataclass for banner, fix tests

Adam-W1:
Separate helpers and filter tests, add dataclass for banner, fix tests

### DIFF
--- a/app/assess/models/round.py
+++ b/app/assess/models/round.py
@@ -30,10 +30,6 @@ class Round:
 
     @staticmethod
     def from_json(data: dict):
-        # eligibility_criteria = {}
-        # if "eligibility_criteria" in data:
-        #     for key, value in data["eligibility_criteria"].items():
-        #         eligibility_criteria.update({key: value})
         return Round(
             title=data.get("title"),
             id=data.get("id"),
--- a/app/assess/routes.py
+++ b/app/assess/routes.py
@@ -10,6 +10,8 @@ from app.assess.forms.continue_application_form import ContinueApplicationForm
 from app.assess.forms.flag_form import FlagApplicationForm
 from app.assess.forms.resolve_flag_form import ResolveFlagForm
 from app.assess.forms.scores_and_justifications import ScoreForm
+from app.assess.helpers import determine_display_status
+from app.assess.helpers import resolve_application
 from app.assess.models.flag import FlagType
 from app.assess.models.ui import applicants_response
 from app.assess.models.ui.assessor_task_list import AssessorTaskList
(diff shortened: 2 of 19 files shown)


app/assess/routes.py
```diff
@@ -117,7 +92,7 @@ def display_sub_criteria(
         "application_id": application_id,
         "fund": fund,
         "comments": comments,
-        "is_flagged": is_flagged,
+        "is_flagged": True if flag else False,
```
gidsg, line 95: can be simplified
```suggestion
        "is_flagged": bool(flag),
```

app/assess/models/round.py
```diff
@@ -30,10 +30,6 @@ def from_dict(cls, d: dict):
 
     @staticmethod
     def from_json(data: dict):
-        # eligibility_criteria = {}
-        # if "eligibility_criteria" in data:
-        #     for key, value in data["eligibility_criteria"].items():
-        #         eligibility_criteria.update({key: value})
```
gidsg, line 36: Can we remove this rather than commenting out?

Adam-W1: that has been removed, it was commented out originally but Im not sure why

gidsg: @Adam-W1 change doesn't seem to have been pushed to remove this?

Adam-W1: The red highlight means it's being removed (Im removing it from main as part of this BAU)

app/assess/views/filters.py
```diff
@@ -8,6 +8,7 @@ def status_to_human(status: str):
         "SUBMITTED": "Submitted",
         "COMPLETED": "Complete",
         "FLAGGED": "Flagged",
+        "STOPPED": "Stopped",
```
gidsg, line 11: Think we could replace this with the built-in Jinja filters either [capitalise](https://tedboy.github.io/jinja2/templ14.html#capitalize) or [title](https://tedboy.github.io/jinja2/templ14.html#title) just awaiting confirmation from design which is correct CC @charlesrt

gidsg: @Adam-W1 design have confirmed `title` case is the intended behaviour, so please can we remove this and use the built-in `title` filter instead?

gidsg: @Adam-W1 So I think on the basis of what design have said. `capitalize` is the correct filter to use.

Adam-W1: I think we still need the custom as the statuses display slightly differently on UI (including removing underscore), For example:
- "NOT_STARTED" -> "Not started"
- "COMPLETED" -> "Assessment completed".

gidsg: Could try doing that chaining a standard filter in the view e.g. `| replace("_"," ")` 

Or could replace this function with 
```
return status.capitalize().replace('_', ' ')
```
and achieve same result.

gidsg: Oh I see missed the assessment complete thing

gidsg (approved):

PR #152, opened 2023-01-24
FS-2178-save-scores-without-rationale

ramsharma-prog:
The "score" macro is created to save any score selected by the user without requiring the user to provide a rationale for the assessment. This means that even if the user does not provide a justification for their selected score, the chosen score will still be saved by the page.


gio-karunakaran (approved):
LGTM

rayhan15243 (approved):

PR #154, opened 2023-01-25
FS-2228-save-radio-button-selection-on-flag-resolve-without-filling-reason

ramsharma-prog:
Saves the radio field option data when clicked  and preselect the radio filed option when page is reloaded


rayhan15243 (approved):

gio-karunakaran (approved):
LGTM

PR #160, opened 2023-01-26
Fs 2174 fixed logout styling

AbuNavsa:


tferns (approved):
Judging by the diff this looks good.  But there are 4 commits, could you squash this for a bit of cleaner history? (when merging)

image

PR #172, opened 2023-02-03
FS-2287: Set unique title tags for pages

gidsg:
Similar approach to https://github.com/communitiesuk/funding-service-design-frontend/pull/243 and https://github.com/communitiesuk/funding-service-design-authenticator/pull/113 to set unique `<title>` tags for pages.

* Pass through theme name to subcriteria template so it is available for title
* Some hacky code is needed to workaround `score` being set by a `theme_id`

### DIFF
--- a/tests/test_routes.py
+++ b/tests/test_routes.py
@@ -4,6 +4,7 @@ import app
 import pytest
 from app.assess.models.flag import Flag
 from app.assess.models.score import Score
+from bs4 import BeautifulSoup
 from config import Config
 from flask import session
 from tests.conftest import create_valid_token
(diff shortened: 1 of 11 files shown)


tests/test_routes.py
```diff
@@ -4,6 +4,7 @@
 import pytest
 from app.assess.models.flag import Flag
 from app.assess.models.score import Score
+from bs4 import BeautifulSoup
```
tferns, line 7: This is actually a really good idea, I'm annoyed I didn't think of this.  Better to use this for assertions in test as opposed to regex.

tferns (approved):

PR #175, opened 2023-02-08
FS-2095: Draft page for "Get help"

tferns:
- Add a route for a help page giving information
- Pull support availability from the fund store
- Edit the logout partial to have "Get help" beside it
- Amend requirements to update fsd-utils

Video demo:
https://user-images.githubusercontent.com/117724519/217511912-17a03bb1-98b6-4516-8e40-11de206fc11d.mp4

**Note:** 
I've left a comment on the JIRA related to this - I couldn't find the styling/markup on the [prototype](https://fsd-pre-award.herokuapp.com/) page, so I was working from a screenshot.  What should be on the blue banner wasn't obvious as it was cropped out of the screenshot, so I just repeated the "Get help" title - in terms of navigation, it was also cropped out, so I went with a sensible "Back" in the same position as the other pages which would simply take the user back to the page they came from.


gidsg:
@tferns I doubt we want the blue banner on this page, but can you check with design to confirm?

tferns:
> @tferns I doubt we want the blue banner on this page, but can you check with design to confirm?

Good point I'll ask in the channel.

tferns:
After speaking to Emma today we've changed the styling a bit, and she has updated the mural.

https://user-images.githubusercontent.com/117724519/217834229-2375e3b0-5d94-4e35-935d-2aaa63a7d1a4.mp4

These changes were pushed in https://github.com/communitiesuk/funding-service-design-assessment/pull/175/commits/6ab022e83cdbe1da40861026dcd84e3058fbc09b

app/default/routes.py
```diff
@@ -23,3 +24,15 @@ def not_found(error):
 @default_bp.errorhandler(500)
 def internal_server_error(error):
     return render_template("500.html"), 500
+
+
+@default_bp.route("/help", methods=["GET"])
```
gidsg, line 29: Could we use `@login_requested` decorator instead i.e. https://github.com/communitiesuk/funding-service-design-utils/blob/bafc22f12d359ccad3dc62ee8830d68bba279e8c/fsd_utils/authentication/decorators.py#L115 which makes auth optional rather than having this in unprotected routes as present?

app/assess/templates/get_help.html
```diff
@@ -0,0 +1,48 @@
...
+
+                {{ logout_partial(sso_logout_url, show_get_help=False) }}
+            </div>
+        </div>
+    {% else %}
+        <div class="fsd-login-banner-background"></div>
+    {% endif %}
+    </div>
+    <div class="govuk-width-container">
+        <h2 class="govuk-heading-l">Get help</h2>
```
gidsg, line 27: Can we make this an `<h1>`? Given I don't _think_ there's an `<h1>` above this on the page?

app/default/routes.py
tferns: From what I can tell `@login_requested` is already called before each request:

https://github.com/communitiesuk/funding-service-design-assessment/blob/6ab022e83cdbe1da40861026dcd84e3058fbc09b/app/create_app.py#L119-L125

So adding `"/help"` was essentially making it an unprotected route, with the change you're suggesting, I don't think we need the `@login_requested` decorator as the `@flask_app.before_request` handles it - I could be misunderstanding something though so let me know if you think I'm missing something.

In short, do we want the `"/help"` route to be protected or not? If not, this PR has that, if we do - I can remove `"help"` from the `unprotected_routes` variable 👍

app/assess/templates/get_help.html
tferns: Good spot, I've changed in https://github.com/communitiesuk/funding-service-design-assessment/pull/175/commits/5e8f874d4caf83c82da3546a2f5de7a84df3a1f7

app/default/routes.py
tferns: Yeah fair point I'll double-check.

tferns: Spoke with Emma again yesterday and decided to make it protected for now, have removed "help" from the unprotected routes in https://github.com/communitiesuk/funding-service-design-assessment/pull/175/commits/909df509571a65bf731b81774764184511c83b6f

tferns: Spoke with Emma again yesterday and decided to make it protected for now, have removed "help" from the unprotected routes in https://github.com/communitiesuk/funding-service-design-assessment/pull/175/commits/909df509571a65bf731b81774764184511c83b6f

app/assess/routes.py
```diff
@@ -29,7 +29,6 @@
 from flask import url_for
 from fsd_utils.authentication.decorators import login_required
 
-
```
gidsg, line 32: Looks like just a whitespace change, can we revert?

gidsg (approved):
Looks good (aside from minor comment)

app/assess/routes.py
tferns: Good point I've re-added in https://github.com/communitiesuk/funding-service-design-assessment/pull/175/commits/e7c94a919f3b27c9a9e796b10a25e0db4cc8c51c, and will squash when merging

PR #176, opened 2023-02-08
FS-2294: refactor score page to own route

gidsg:
Previously we rendered the scoring pages by overloading the `theme_id` parameter with `score`. This caused a lot of hacky workarounds and branching in templates to ensure the correct things were rendered and meant `theme_id` couldn't be safely used semantically i.e. to lookup a `Theme` class.

This moves all scoring to its own route `/application_id/<application_id>/sub_criteria_id/<sub_criteria_id>/score`

This is purely a refactor, there should be no change in any page functionality.

### DIFF
--- a/app/assess/routes.py
+++ b/app/assess/routes.py
@@ -194,10 +194,18 @@ def display_sub_criteria(
     "/application_id/<application_id>/sub_criteria_id/<sub_criteria_id>/score",
     methods=["POST", "GET"],
 )
+@login_required(roles_required=["LEAD_ASSESSOR", "ASSESSOR"])
 def score(
     application_id,
     sub_criteria_id,
 ):
+    sub_criteria: SubCriteria = get_sub_criteria(
+        application_id, sub_criteria_id
+    )
+
+    if not sub_criteria.is_scored:
+        abort(404)
+
     return "IN HERE"
 
 
--- a/app/assess/routes.py
+++ b/app/assess/routes.py
@@ -191,6 +191,17 @@ def display_sub_criteria(
 
 
 @assess_bp.route(
+    "/application_id/<application_id>/sub_criteria_id/<sub_criteria_id>/score",
+    methods=["POST", "GET"],
+)
+def score(
+    application_id,
+    sub_criteria_id,
+):
+    return "IN HERE"
+
+
+@assess_bp.route(
     "/flag/<application_id>",
     methods=["GET", "POST"],
 )
--- a/app/assess/routes.py
+++ b/app/assess/routes.py
@@ -165,7 +165,7 @@ def display_sub_criteria(
         ]
         return render_template(
             "sub_criteria.html",
-            on_summary=True,
+            is_score_page=True,
             score_list=score_list or None,
             latest_score=latest_score,
             COF_score_list=COF_score_list,
@@ -184,7 +184,7 @@ def display_sub_criteria(
 
     return render_template(
         "sub_criteria.html",
-        on_summary=False,
+        is_score_page=False,
         answers_meta=answers_meta,
         **common_template_config,
     )
@@ -256,7 +256,7 @@ def score(
     return render_template(
         "score.html",
         application_id=application_id,
-        on_summary=True,
+        is_score_page=True,
         score_list=score_list or None,
         latest_score=latest_score,
         COF_score_list=COF_score_list,
--- a/app/assess/routes.py
+++ b/app/assess/routes.py
@@ -255,6 +255,7 @@ def score(
     ]
     return render_template(
         "score.html",
+        application_id=application_id,
         on_summary=True,
         score_list=score_list or None,
         latest_score=latest_score,
--- a/app/assess/routes.py
+++ b/app/assess/routes.py
@@ -133,7 +133,7 @@ def score(
 
     if not sub_criteria.is_scored:
         abort(404)
-    theme_id = request.args.get(sub_criteria.themes[0].id)
+    theme_id = sub_criteria.themes[0].id
     fund = get_fund(Config.COF_FUND_ID)
     flag = get_latest_flag(application_id)
     comments = get_comments(
--- a/app/assess/routes.py
+++ b/app/assess/routes.py
@@ -133,13 +133,15 @@ def score(
 
     if not sub_criteria.is_scored:
         abort(404)
-    theme_id = sub_criteria.themes[0].id
     fund = get_fund(Config.COF_FUND_ID)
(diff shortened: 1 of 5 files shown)


app/assess/routes.py
```diff
@@ -184,12 +113,89 @@ def display_sub_criteria(
...
+    sub_criteria: SubCriteria = get_sub_criteria(
+        application_id, sub_criteria_id
+    )
+
+    if not sub_criteria.is_scored:
+        abort(404)
+    theme_id = sub_criteria.themes[0].id
+    fund = get_fund(Config.COF_FUND_ID)
+    flag = get_latest_flag(application_id)
+    comments = get_comments(
```
Adam-W1, line 139: is the way we retrieve comments here correct for the scoring route?

Adam-W1: To return comments for a score it looks like the assessment store is set up to retrieve 'theme_id = "score' https://github.com/communitiesuk/funding-service-design-assessment-store/blob/main/db/queries/comments/queries.py#:~:text=if%20theme_id%20%3D%3D%20%22score%22%3A

app/assess/routes.py
```diff
@@ -184,12 +113,89 @@ def display_sub_criteria(
...
+def score(
+    application_id,
+    sub_criteria_id,
+):
+    sub_criteria: SubCriteria = get_sub_criteria(
+        application_id, sub_criteria_id
+    )
+
+    if not sub_criteria.is_scored:
+        abort(404)
```
tferns, line 135: I like this early exit

app/assess/templates/macros/sub_criteria_navbar.html
```diff
@@ -10,9 +9,9 @@
     </ul>
 
 {% if g.user.highest_role in ["ASSESSOR","LEAD_ASSESSOR"] %}
-    {% if current_theme_id != "score" and sub_criteria.is_scored == True %}
+    {% if not is_score_page and sub_criteria.is_scored %}
     <ul class="govuk-list">
-        <li><a id="score-subcriteria-link" class="{{'govuk-link--no-underline govuk-list' if on_summary else 'govuk-link--no-visited-state'}}" href="{{ url_for('assess_bp.display_sub_criteria',application_id=application_id,sub_criteria_id=sub_criteria.id,theme_id='score') }}">Score the subcriteria</a></li>
+        <li><a id="score-subcriteria-link" class="{{'govuk-link--no-underline govuk-list' if is_score_page else 'govuk-link--no-visited-state'}}" href="{{ url_for('assess_bp.score',application_id=application_id,sub_criteria_id=sub_criteria.id) }}">Score the subcriteria</a></li>
```
tferns, line 14: The `if is_score_page` is redundant as based on L12, we only render this block given it is `False`, so it will always go to the `else` statement

tferns (commented):
Looks good other than what Adam mentioned, think we might have an issue with the query pulling back comments - although the default of theme_id is `None` so maybe that pulls back all comments?... Not sure.

app/assess/routes.py
gidsg: @Adam-W1 Good spot 👀 
Sad we have included this concept in the API too 😢 
Have fixed and added TODO in 023370a.

app/assess/templates/macros/sub_criteria_navbar.html
gidsg: @tferns Although it is defaulted in the macro, we can still override it by setting the argument which we do in https://github.com/communitiesuk/funding-service-design-assessment/blob/023370a1816aa6988f033ed79e8bf36c07a5bb45/app/assess/templates/score.html#L43

tferns: Ah, I was referring to the encapsulating if statement, i.e:
https://github.com/communitiesuk/funding-service-design-assessment/blob/194bff396de7e89d1f088908c30154e01f3d21be/app/assess/templates/macros/sub_criteria_navbar.html#L12

This condition needs to be met for the `<ul...` to be rendered, so the nested if condition is never met:
https://github.com/communitiesuk/funding-service-design-assessment/blob/194bff396de7e89d1f088908c30154e01f3d21be/app/assess/templates/macros/sub_criteria_navbar.html#L14

i.e
```python3
is_not_true = False
if not is_not_true:
    if is_not_true: # will never be called
```

tferns: Ah, I was referring to the encapsulating if statement:
https://github.com/communitiesuk/funding-service-design-assessment/blob/194bff396de7e89d1f088908c30154e01f3d21be/app/assess/templates/macros/sub_criteria_navbar.html#L12

This condition needs to be met for the `<ul...` to be rendered, so the nested if condition is never met:
https://github.com/communitiesuk/funding-service-design-assessment/blob/194bff396de7e89d1f088908c30154e01f3d21be/app/assess/templates/macros/sub_criteria_navbar.html#L14

i.e
```python3
is_not_true = False
if not is_not_true:
    if is_not_true: # will never be called
```

tferns: Ah, I was referring to the encapsulating if statement:
https://github.com/communitiesuk/funding-service-design-assessment/blob/194bff396de7e89d1f088908c30154e01f3d21be/app/assess/templates/macros/sub_criteria_navbar.html#L12

This condition needs to be met for the `<ul...` to be rendered, so the nested if condition is never met:
https://github.com/communitiesuk/funding-service-design-assessment/blob/194bff396de7e89d1f088908c30154e01f3d21be/app/assess/templates/macros/sub_criteria_navbar.html#L14

i.e
```python3
is_not_true = False
if not is_not_true:
    if is_not_true: # will never be called
```

tferns:
```suggestion
        <li><a id="score-subcriteria-link" class="govuk-link--no-visited-state" href="{{ url_for('assess_bp.score',application_id=application_id,sub_criteria_id=sub_criteria.id) }}">Score the subcriteria</a></li>
```

gidsg: @tferns ah yes, I agree the code is unreachable but I believe the same bug is present as-is in `main` e.g. https://assessment.uat.gids.dev/assess/application_id/0b3e5a51-e9a1-4000-9d27-79e111edf737/sub_criteria_id/benefits?theme_id=score so would it be OK to fix this in a separate PR, rather than changing the observed behaviour in this one?

tferns (approved):

PR #179, opened 2023-02-10
FS-2272: don't pass theme_id=score to comment endpoint on score page

gidsg:
Now that https://github.com/communitiesuk/funding-service-design-assessment-store/pull/123 has been merged and deployed we no longer need to pass `theme_id=score` to the comments endpoint.

This also refactors how the URL is built so we can strip the parameter from the generated query string if `None`.

### DIFF
--- a/app/assess/data.py
+++ b/app/assess/data.py
@@ -432,22 +432,31 @@ def get_sub_criteria_theme_answers(
         abort(404, description=msg)
 
 
-def get_comments(application_id: str, sub_criteria_id: str, theme_id, themes):
-    """_summary_: Function is set up to retrieve
-    the data from application store with
-    get_data() function.
+def get_comments(
+    application_id: str, sub_criteria_id: str, theme_id: str | None, themes
+):
+    """_summary_: Get comments from the assessment store
     Args:
-        application_id: Takes an application_id, sub_criteria_id: takes a sub_criteria_id # noqa
+        application_id: application_id,
+        sub_criteria_id: sub_criteria_id
+        theme_id: optional theme_id (else returns all comments for subcriteria)
+        themes: list of subcriteria themes
     Returns:
         Returns a dictionary of comments.
     """
+    query_params = {
+        "application_id": application_id,
+        "sub_criteria_id": sub_criteria_id,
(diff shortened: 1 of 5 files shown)


tests/test_routes.py
```diff
@@ -205,6 +205,7 @@ def test_route_sub_criteria_scoring(
                 soup.title.string
                 == "Score - Engagement - Community Gym - Assessment Hub"
             )
+            assert b"This is a comment" in response.data
```
Adam-W1, line 208: Not part of this PR but it reminds me of @srh-sloan [post](https://communities-govuk.slack.com/archives/C02PHBER287/p1675866903115839) about testing templates, this looks like another approach 
[here](https://github.com/communitiesuk/funding-service-design-assessment/blob/1b63c296082702b2a73361d5d23b1a00d221529b/tests/test_routes.py#L645:~:text=assert%201%20%3D%3D-,len,-(templates_rendered))

Adam-W1: I'm creating a page for standardised testing approaches that people can use to guide them in writing tests on our repos going forward, maybe we can out a chat in next week with others and put something concrete in the guidance?

app/assess/data.py
```diff
@@ -432,22 +432,31 @@ def get_sub_criteria_theme_answers(
...
+    }
     comment_endpoint = (
         Config.ASSESSMENT_STORE_API_HOST
-        + Config.COMMENTS_ENDPOINT.format(
-            application_id=application_id,
-            sub_criteria_id=sub_criteria_id,
-            theme_id=theme_id,
-        )
+        + "/comment?"
+        + urlencode(query=query_params_strip_nones)
```
Adam-W1, line 459: The "ASSESSMENT_STORE/comment" is in the env var 'ASSESSMENT_COMMENT_ENDPOINT' (in the default config). 

Also, nice use of urlencode 👍 Ive not seen it in other places so one to add to tech debt.

gidsg: @Adam-W1 have updated now to use this variable in f71ff14

Adam-W1 (approved):
👌

PR #180, opened 2023-02-15
FS-2292: remove unused files

gidsg:
Presume these were part of an early spike but none of them appears to be used.

I have done some manual tests and run unit tests but we should also run e2e tests to confirm this causes no regressions.


adamdavies1 (approved):
LGTM

PR #182, opened 2023-02-17
FS-2342 display status

gidsg:
### Change description
Previous `determine_display_status` was a side-effecting function which mutated data classes to add two properties `display_status` and `flag_resoved` to classes. This made it very difficult to test and reason about.

This refactors in a couple of ways:
* move `flag_resolved` into an `is_flaggable` variable to reflect its usage and pass to template where used
* make `determine_display_status` a simpler function that returns a value and pass that to the templates
* Adds tests for `determine_display_status`

### How to test
See unit tests


app/assess/routes.py
```diff
@@ -143,7 +147,7 @@ def score(
     )
     # TODO: Refactor this function so it doesn't rely on side-effects
     # and mutating SubCriteria model
-    determine_display_status(sub_criteria, flag)
+    determine_display_status(sub_criteria.workflow_status, flag)
```
tferns, line 150: Do we need to do something with the return value here? And could we remove the TODO comment?

Adam-W1, line 150: Could we assign to display status here also?

tferns (commented):
Nice refactor, looks good but I think you may have missed one usage

app/assess/routes.py
```diff
@@ -372,8 +376,8 @@ def application(application_id):
             abort(404)
         assessor_task_list_metadata["fund_name"] = fund.name
         state = AssessorTaskList.from_json(assessor_task_list_metadata)
-
-    determine_display_status(state, flag)
+    is_flaggable = not flag or flag.flag_type == FlagType.RESOLVED
```
Adam-W1, line 379: Might this be
 flag.flag_type in (FlagType.RESOLVED, FlagType.QA_COMPLETE)
as we can still flag qa complete applications

Adam-W1 (commented):
Looks neater 👍  Just a couple comments

app/assess/routes.py
gidsg: Good spot, fixed in 85ee0e9

gidsg: Good spot fixed in 85ee0e9

app/assess/routes.py
gidsg: Good spot, fixed in 85ee0e9

Adam-W1 (approved):
👌

PR #187, opened 2023-02-22
Fs 2303 error summary scores comments

AbuNavsa:
FS-2303 Display error summary banner
### Description
Added the banner to display the error summary when score inputs aren't filled. Banner has also been added for when comment box is not filled.

### Screenshots of UI changes
image

image

### DIFF
--- a/app/assess/templates/sub_criteria.html
+++ b/app/assess/templates/sub_criteria.html
@@ -50,6 +50,23 @@
                 </div>
             {% endif %}
     {% endif %}
+
+    {% if comment_form.comment.errors %}
+        <div class="govuk-error-summary" data-module="govuk-error-summary">
+            <div role="alert">
+            <h2 class="govuk-error-summary__title">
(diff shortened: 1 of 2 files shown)


app/assess/templates/sub_criteria.html
```diff
@@ -27,8 +27,53 @@
         </p>
 
         {{ logout_partial(sso_logout_url) }}
+
     </div>
 
+    {% if current_theme_id == 'score' %}
```
gidsg, line 33: We aren't using this parameter anymore, I think this logic should be in the `score.html` e.g. https://github.com/communitiesuk/funding-service-design-assessment/blob/f9269f74c6728cb2348296dd58cea63ee7e3c1ff/app/assess/routes.py#L188-L202

gidsg (changes requested):

gidsg (approved):

PR #188, opened 2023-02-22
FS-2371: Fix for duplicate ids in bulk-accounts endpoint

tferns:
### Change description

When making the request for bulk-accounts, we unnecessarily send the same ids lots of times, by putting it into a set then back to a list, we can ensure each id is only sent once.

```python
list(set(['a', 'a', 'b'])) # ['b', 'a']
```

### How to test

N/A - This is a quick tiny fix to delay the issue from re-occurring on from e2e tests, but in theory it could still happen if lots of unique users were to comment on the same application, so I think it would be good to change it from using query parameters based on that, although that's a lot less likely to occur after we de-duplicate the ids.   More info on the [JIRA](https://digital.dclg.gov.uk/jira/browse/FS-2371).

### DIFF
--- a/app/assess/data.py
+++ b/app/assess/data.py
@@ -180,7 +180,7 @@ def get_round_with_applications(
 
 def get_bulk_accounts_dict(account_ids: List):
     account_url = Config.BULK_ACCOUNTS_ENDPOINT
-    account_params = {"account_id": account_ids}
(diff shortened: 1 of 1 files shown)


rayhan15243 (approved):

gidsg:
@tferns long-term we should use a POST request for this, as described in the JIRA ticket.

PR #200, opened 2023-03-10
FS-2415-Standardize-Workflows

gio-karunakaran:
Update: Improve use of shared workflows across FSD repositories

This PR aims to enhance the use of shared workflows in the FSD organization by updating our approach to repository-specific workflows. Currently, not all repositories use the shared workflows defined in the [workflows repository](https://github.com/communitiesuk/funding-service-design-workflows/blob/main/.github/workflows/deploy.yml), resulting in duplicative and inconsistent processes across the organization.

To address this, we propose utilizing a single job that points to the deploy.yml file in the workflows repository and sets variables/flags and validates secrets based on the type of service. This will streamline our workflow processes and ensure consistency across our repos.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -38,7 +38,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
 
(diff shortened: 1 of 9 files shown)


rayhan15243 (approved):
Looks good to me.

PR #222, opened 2023-04-12
FS-2561-bug-fix-when-scoring-subcriteria-for-first-time

ramsharma-prog:
This PR fixes the bug when assessor/Lead assessor wants to score a sub criteria for the first time it returns 500 internal server error.

### DIFF
--- a/app/assess/data.py
+++ b/app/assess/data.py
@@ -140,8 +140,9 @@ def get_round(fund_id: str, round_id: str) -> Union[Round, None]:
 
 def get_bulk_accounts_dict(account_ids: List):
     account_url = Config.BULK_ACCOUNTS_ENDPOINT
-    account_params = {"account_id": list(set(account_ids))}
-    return get_data(account_url, account_params)
+    if account_ids:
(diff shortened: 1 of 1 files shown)


rayhan15243 (approved):

rayhan15243 (approved):

app/assess/data.py
```diff
@@ -140,8 +140,9 @@ def get_round(fund_id: str, round_id: str) -> Union[Round, None]:
 
 def get_bulk_accounts_dict(account_ids: List):
     account_url = Config.BULK_ACCOUNTS_ENDPOINT
-    account_params = {"account_id": list(set(account_ids))}
-    return get_data(account_url, account_params)
+    if account_ids:
```
Adam-W1, line 143: It might be useful to log a warn here if its called with no account ids

ramsharma-prog: Thank you for spotting. Logs added

rayhan15243 (dismissed):

gio-karunakaran (approved):
👍🏾

rayhan15243 (approved):

PR #226, opened 2023-04-17
FS-2585-display-assessment-service-time-in-bst

ramsharma-prog:
This PR adds a new function utc_to_bst that converts a datetime string in UTC format to BST timezone and returns a formatted string. The function supports two datetime formats:

["%Y-%m-%dT%H:%M:%S.%f",
 "%Y-%m-%d %H:%M:%S.%f", 
"%Y-%m-%d %H:%M:%S"] 

and raises a ValueError if the input datetime string does not match either format.


app/assess/views/filters.py
```diff
@@ -53,3 +53,18 @@ def all_caps_to_human(word):
...
+        try:
+            utc_time = datetime.strptime(value, dt_format)
+            break
+        except ValueError:
+            pass
+    else:
+        raise ValueError("Invalid datetime format")
+
+    uk_time = utc_time.astimezone(timezone(tz))
+    return uk_time.strftime("%d/%m/%Y at %H:%M")
```
tferns, line 70: Can we modify the existing function instead of adding a new one?

https://github.com/communitiesuk/funding-service-design-assessment/blob/fe0e55e250bbfc5005d096707aae23f24b19aa2d/app/assess/views/filters.py#L23-L26

ramsharma-prog: I renamed the utc_to_tz function to utc_to_bst because the new name more accurately describes what the function does?

tferns: Nice sounds good!  I'd called it `utc_to_tz` as the second parameter could optionally take another timezone, but I don't think we'll ever need that so don't mind `utc_to_btc` - we could potentially remove the `tz` option from the method signature all together and just have "Europe/London" hard-coded? Then the function name would be most accurate I think

tferns (approved):

rayhan15243 (approved):

PR #227, opened 2023-04-19
enable caching for static assets

RamuniN:
JIRA link
https://digital.dclg.gov.uk/jira/browse/FS-2587

### Change description
Fixed the page flickering by enabling the caching only on static assets.

- [x] Unit tests and other appropriate tests added or updated
- [x] README and other documentation has been updated / added (if needed)

### How to test
Just click on any link & notice the disappeared page flickering.

### DIFF
--- a/app/create_app.py
+++ b/app/create_app.py
@@ -114,7 +114,7 @@ def create_app() -> Flask:
             )
 
         static_files_list = []
-        for _, _, files in os.walk(os.getcwd() + "/app/static"):
+        for _, _, files in os.walk(flask_app.static_url_path):
             for file in files:
                 # append the file name to the list
                 static_files_list.append(file)
--- a/app/create_app.py
+++ b/app/create_app.py
@@ -113,10 +113,10 @@ def create_app() -> Flask:
                 unprotected_routes=["/", "/healthcheck"],
             )
 
+        # Get static filenames list
         static_files_list = []
         for _, _, files in os.walk(flask_app.static_url_path):
             for file in files:
-                # append the file name to the list
                 static_files_list.append(file)
 
         @flask_app.after_request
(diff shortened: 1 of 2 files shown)


app/create_app.py
```diff
@@ -110,13 +113,27 @@ def ensure_minimum_required_roles():
                 unprotected_routes=["/", "/healthcheck"],
             )
 
+        static_files_list = []
+        for _, _, files in os.walk(os.getcwd() + "/app/static"):
```
RamuniN, line 117: want to use `STATIC_URL_PATH` here instead of hardcoded. But the value set to `STATIC_URL_PATH = "/assets"` in config/default.py which seems to be wrong

Adam-W1: we could change this location maybe?

RamuniN: done

app/create_app.py
```diff
@@ -110,13 +113,31 @@ def ensure_minimum_required_roles():
...
         @flask_app.after_request
         def set_response_headers(response):
-            response.headers[
-                "Cache-Control"
-            ] = "no-cache, no-store, must-revalidate"
-            response.headers["Pragma"] = "no-cache"
-            response.headers["Expires"] = "0"
+            filename = (
+                re.findall(
+                    "filename=(.+)", response.headers["Content-Disposition"]
```
Adam-W1, line 126: nice use of re. Should we be checking for "inline" only here?

Adam-W1 (commented):
Looks great Navya - nice fix. Just the one small comment

app/create_app.py
RamuniN: yes. ```
response.headers.get("Content-Disposition") = 
'inline; filename=govuk-frontend-4.0.0.min.css'
```
trying extract filename from here

RamuniN: I think it's a good point, but should we cache when it's "attachment" also ?

Adam-W1 (approved):

app/create_app.py
Adam-W1: yeah, i was just checking but i think it makes sense to cache both to be fair

PR #232, opened 2023-04-28
FS-2640-location-missing-for-some-applications

ramsharma-prog:
This PR contains a logic to replace the default location name with a desired name.


gio-karunakaran (approved):
LGTM

PR #263, opened 2023-06-15
BAU: Fix copy styles

tferns:
Teeny tiny PR, means we copy the whole folder instead of just styles.

There are `scss` & `js` that need copied, so bringing it a layer down picks them up too.


srh-sloan (approved):
Nice 👍

PR #270, opened 2023-06-21
BAU: Add safe-guard for mock assessments

tferns:
We've seeded some mock assessments for NSTF, these don't have files attached as they haven't went through the full end-to-end journey, so when we are attempting to retrieve the files from AWS, we're getting an error since the keys don't exist.

This PR adds a small safeguard for that scenario, so we don't explode the page when this happens.


adamdavies1 (approved):
LGTM

PR #273, opened 2023-06-21
FS-3020 Remove references to cof

adamdavies1:
[FS-3020 Refactor explicit references to COF in assessment code to general fund](https://digital.dclg.gov.uk/jira/browse/FS-3020)

### Change description
Removed references to COF and Fixed issue with old community fund displaying on the sub criteria page

### How to test
Got to the sub criteria page for night shelter and make sure that it banner says Night Shelter

image

### DIFF
--- a/app/assess/routes.py
+++ b/app/assess/routes.py
@@ -98,7 +98,6 @@ def display_sub_criteria(
     if not assessor_task_list_metadata:
         abort(404)
 
-    # maybe there's a better way to do this?..
     fund = get_fund(assessor_task_list_metadata["fund_id"])
     flag = get_latest_flag(application_id)
 
@@ -166,7 +165,6 @@ def score(
     if not assessor_task_list_metadata:
         abort(404)
 
-    # maybe there's a better way to do this?..
     fund = get_fund(assessor_task_list_metadata["fund_id"])
     flag = get_latest_flag(application_id)
 
--- a/app/assess/routes.py
+++ b/app/assess/routes.py
@@ -93,7 +93,13 @@ def display_sub_criteria(
             )
         )
 
-    fund = get_fund(Config.COF_FUND_ID)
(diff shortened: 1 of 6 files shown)


app/assess/routes.py
```diff
@@ -93,7 +93,12 @@ def display_sub_criteria(
             )
         )
 
-    fund = get_fund(Config.COF_FUND_ID)
+    assessor_task_list_metadata = get_assessor_task_list_state(application_id)
```
adamdavies1, line 96: Im wondering if we need to make a get fund method thats takes the application id instead of having you go to get_assessor_task_list_state

srh-sloan: We have the fund and round IDs when we load the dashboard page (eg. assessor_dashboard/nstf/r2) so could we just pass these in thru hidden form fields when we load the other routes and then use those to directly retrieve the fund name when we need it?

srh-sloan (commented):
All fine but as per your comment, could you just use this [get_fund](https://github.com/communitiesuk/funding-service-design-assessment/blob/46bd398322baedf1fe3faa35520a7b6546a031e8/app/assess/data.py#L104) function to retrieve the fund name if you pass the fund/round id round thru the different routes? Instead of loading full application metadata as you say.

tferns (approved):

app/assess/routes.py
tferns: Hmm, I don’t know if it’s worth worrying about this too much imo.  It’s just a bit of data and a get request, personal preference would be to have it reliable and encapsulated behind the scenes - if it becomes a performance issue in the future (which I doubt) then we can maybe change it then? 

I don’t think using a hidden field would work for GET requests, and even then it seems a bit off having to use a hidden field just to pass some data, users could also malform that if passed in through the UI form too.

PR #285, opened 2023-06-30
BAU: Allow all access when forcing funds accessible

tferns:
- Add an all roles access controller when we're forcing all funds as accessible


adamdavies1 (approved):
LGTM

PR #286, opened 2023-07-03
BAU: Fix bulk accounts functionality

tferns:
Issue:

 `users_result` is a dictionary, and we were previously using it as a single object - this should fix the issues we are seeing on assessment at the moment, an oversight on my part.

### DIFF
--- a/app/assess/data.py
+++ b/app/assess/data.py
@@ -162,11 +162,11 @@ def get_bulk_accounts_dict(account_ids: List, fund_short_name: str):
             del debug_user_config["highest_role"]
             users_result[Config.DEBUG_USER_ACCOUNT_ID] = debug_user_config
 
-        # we only need the highest role for the fund we are currently viewing
-        users_result["highest_role"] = users_result["highest_role_map"][
-            fund_short_name
-        ]
-        del users_result["highest_role_map"]
+        for user_result in users_result.values():
+            # we only need the highest role for the fund we are currently viewing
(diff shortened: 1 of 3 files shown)


app/assess/data.py
```diff
@@ -162,11 +162,11 @@ def get_bulk_accounts_dict(account_ids: List, fund_short_name: str):
             del debug_user_config["highest_role"]
```
RamuniN, line 162:
```suggestion
            debug_user_config["highest_role_map"] = {fund_short_name: fund_short_name + "_LEAD_ASSESSOR"}
            del debug_user_config["highest_role"]
```

RamuniN: can we make DEBUG_USER_ROLE dynamic. suggestion to replace Line 161

RamuniN:
can you also add 
```
            "NSTF_LEAD_ASSESSOR",
            "NSTF_ASSESSOR",
            "NSTF_COMMENTER",
```
[here](https://github.com/communitiesuk/funding-service-design-assessment/blob/e1f6cb6c49b45bf3095bf0fa0c0527e56eb96dbc/config/envs/development.py#L31)

app/assess/data.py
tferns: I see what you mean.  I think it would be `{"COF": "LEAD_ASSESSOR"}` as `highest_role` is used only for presentation now afaik.  So doesn't require the `COF_` or `NSTF_` prefix.

RamuniN (approved):
👍

PR #287, opened 2023-07-03
fix debug role settings

RamuniN:
- fix bulk bulk accounts stuff
- remove Todo code (flag history)

### DIFF
--- a/app/assess/data.py
+++ b/app/assess/data.py
@@ -158,15 +158,17 @@ def get_bulk_accounts_dict(account_ids: List, fund_short_name: str):
         if Config.FLASK_ENV == "development":
             debug_user_config = deepcopy(Config.DEBUG_USER)
             debug_user_config["email_address"] = Config.DEBUG_USER["email"]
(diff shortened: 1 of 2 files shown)


app/assess/data.py
```diff
@@ -158,15 +158,17 @@ def get_bulk_accounts_dict(account_ids: List, fund_short_name: str):
         if Config.FLASK_ENV == "development":
             debug_user_config = deepcopy(Config.DEBUG_USER)
             debug_user_config["email_address"] = Config.DEBUG_USER["email"]
-            debug_user_config["highest_role_map"] = {"COF": "LEAD_ASSESSOR"}
+            debug_user_config["highest_role_map"] = {
+                fund_short_name: fund_short_name + "_" + Config.DEBUG_USER_ROLE
+            }
```
tferns, line 163: I don't think the `highest_role` would have a prefix, as it's only used as presentation on the UI, we could have the prefix if we wanted to show on the UI i.e `COF Lead Assessor` or `NSTF Lead Assessor` - but I imagine we may not want short names there and instead `Community Ownership Fund Lead Assessor`?

tferns (approved):

PR #290, opened 2023-07-04
BAU: Change dict to .get so we safely retrieve roles

tferns:
- Instead of getting roles by using the key `[key]` - we instead use `.get()` with a default of an empty string
- This can happen if a user is un-assigned all their roles, but had left comments/scores/flags etc in the past


gio-karunakaran (approved):
👍🏾

PR #291, opened 2023-07-04
FS-3069: Lead assessor export data

tferns:
https://dluhcdigital.atlassian.net/browse/FS-3069

`is_lead_assessor` doesn't exist, we need to check `g.access_controller.is_lead_assessor` to fix the regression


gio-karunakaran (approved):
👍🏾

PR #294, opened 2023-07-04
BAU: Remove force environment variables

tferns:
This code was used temporarily to force the funds open and disable devolved authority auth, just removing it now.


gio-karunakaran (approved):
👍🏾

PR #313, opened 2023-07-10
bau/bug-fix-highest-role-switched-to-highest-role-map

ramsharma-prog:
This PR includes a bug fix for the User attribute "highest_role" and updates the utils version to 2.0.10, which enhances text file formatting and addresses a bug.


Adam-W1 (approved):

app/assess/routes.py
```diff
@@ -557,7 +557,7 @@ def application(application_id):
         accounts_list=accounts_list,
         teams_flag_stats=teams_flag_stats,
         flags_list=flags_list,
-        current_user_role=g.user.highest_role,
+        current_user_role=g.user.highest_role_map,
```
tferns, line 560: This isn't used anymore, think it's better to remove it?

tests/test_jinja_macros.py
```diff
@@ -41,7 +41,7 @@ def default_flask_g():
         full_name="Test Lead Assessor",
         email="test@example.com",
         roles=["COF_LEAD_ASSESSOR", "COF_ASSESSOR", "COF_COMMENTER"],
-        highest_role="COF_LEAD_ASSESSOR",
+        highest_role_map="COF_LEAD_ASSESSOR",
```
tferns, line 44: This should maybe be a map, but it's not used anymore so might be able to remove the g.user from this test all together?

tests/test_jinja_macros.py
```diff
@@ -119,7 +119,7 @@ def test_criteria_macro_commenter(self, request_ctx):
             full_name="Test Commenter",
             email="test@example.com",
             roles=["COF_COMMENTER"],
-            highest_role="COF_COMMENTER",
+            highest_role_map="COF_COMMENTER",
```
tferns, line 122: This should maybe be a map, but it's not used anymore so might be able to remove the g.user from this test all together?

tests/test_jinja_macros.py
ramsharma-prog: User class has an attribute "highest_role_map" in utils?

tests/test_jinja_macros.py
ramsharma-prog: User class has an attribute "highest_role_map" in utils?

app/assess/routes.py
tferns:
```suggestion
```

tests/test_jinja_macros.py
ramsharma-prog: I would rather have a conversation with the person who has been working on this functionality but my focus is more on the functionality I have been working on related to mapping and formatting.

tests/test_jinja_macros.py
ramsharma-prog: I would rather have a conversation with the person who has been working on this functionality but my focus is more on the functionality I have been working on related to mapping and formatting.

app/assess/routes.py
ramsharma-prog: removed

tferns (approved):

PR #333, opened 2023-07-17
fix type hint in function return

RamuniN:
https://dluhcdigital.atlassian.net/browse/FS-3215

### Description
fix type hint in function return

### DIFF
--- a/app/assess/data.py
+++ b/app/assess/data.py
@@ -401,25 +401,16 @@ def get_flags(application_id: str) -> List[FlagV2]:
     if flag:
         return FlagV2.from_list(flag)
     else:
-        msg = f"flag for application: '{application_id}' not found."
-        current_app.logger.warn(msg)
         return []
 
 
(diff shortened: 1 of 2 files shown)


app/assess/data.py
```diff
@@ -401,25 +401,16 @@ def get_flags(application_id: str) -> List[FlagV2]:
...
         )
     )
-    if qa_complete:
-        return qa_complete
-    else:
-        msg = (
-            f"qa_complete info for application: '{application_id}' not found."
-        )
-        current_app.logger.warn(msg)
-        return {}
```
gio-karunakaran, line 422: do we want to remove this?

gio-karunakaran (commented):
just a couple Qs

app/assess/data.py
RamuniN: Line 408 already returns empty dict when QA is not completed. so no need of duplicating code and log the warning messages

PR #335, opened 2023-07-17
BAU: Add 403 forbidden template and handler

tferns:
- Add page for 403 forbidden, since it's now a common use case
- If someone without a Wales role is linked a Wales application, they may see a 403, for example
- Also if someone without a COF role is linked a COF application, they may see a 403


RamuniN (approved):
LGTM.

PR #337, opened 2023-07-18
Makes sure valid team_name is available before rendering

AbuNavsa:
FS-3233: Flag application COF - Applications with no allocation team assigned displays separate metrics

### Change description
Fixed bug where old data, so applications that were flagged without a team allocation, would render a stats box with the team name reading as None. Now only renders if team_name is truthy


RamuniN (approved):
👍

PR #368, opened 2023-07-31
Adjusting layout spacing and widths to improve readability

stevewalton28:
Adjusting some widths and margins to improve layout in assessment.
- Increasing max-width containers by 200px from 1400px to 1600px
- Increasing max-width on dashobard views for assessors/commentors to 95% view witdth

This will have no impact on those on smaller screens (their expereince will be the same).  For those on larger screens or using external monitors, the experience should improve as more details are added to the dashboard views. This change is a precursor to suggestions made from Product teams for additional dashboard views/customisable columns.

Adjusting layout of previously "floating" buttons for flagging and marking QA complete so they are no longer floating and are in line.

Adding hint/warning text when there is no flag history to display.

Fixing govuk secondary button colors (previously displaying with no background color).

Changes agreed with Sudheer/deisng at Leads mtg today.


tferns (approved):
lgtm 👍

PR #374, opened 2023-08-04
FIX: protect link and route with auth

Adam-W1:


rayhan15243 (approved):
Splendid work as always @Adam-W1

PR #377, opened 2023-08-07
Improve performance of assessment_dashboard

RamuniN:
https://dluhcdigital.atlassian.net/browse/FS-3269

## Description
Current implementation makes multiple API calls to tagging endpoint for every application which caused the performance issues on assessment dashboard. 
- Tags are now retrieved with single API `application overviews endpoint`(no need to make separate API call to fetch tags) to improve the performance
- updated the mock data & unit tests

Also check the PR https://github.com/communitiesuk/funding-service-design-assessment-store/pull/261

### Before (3.94 seconds to load 50 applications)
image
### After (1.73 seconds to load 50 applications)
image

### DIFF
--- a/app/assess/helpers.py
+++ b/app/assess/helpers.py
@@ -1,4 +1,3 @@
-import concurrent
 import csv
 import time
 from collections import OrderedDict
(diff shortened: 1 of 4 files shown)


Adam-W1 (approved):
lgtm

PR #384, opened 2023-08-11
FS-3235: Format answers for csv/pdf & expose "pdf" on UI

tferns:
- Adds all file types to the UI on the export page
- Uses existing `format_answers` function on answers on csv/pdf


adamdavies1 (approved):
LGTM

PR #432, opened 2023-10-03
FS-3339 added lru cache for fund store endpoints

RamuniN:
https://dluhcdigital.atlassian.net/browse/FS-3339

### Description
- Implemented lru caching for fund store endpoints
- Updated & added unittests
- Cache time is set as 3600s on prod, 300s on test & 30s on dev

As initially thought, implementing LRU cache on assessment store endpoints is not the solution for performance improvement. As any caching on the assessment store endpoint will not reflect changes made by assessors on flagging or tagging immediately. So caching is added only for fund-store endpoints in this PR. Please look at the ticket comment in Jira for more information.

### How to test
- Put breakpoints in the get_fund and get_round methods of fund_store
- Load the assessment frontend in a browser and check that the above breakpoints don't get hit except on the first retrieval of a session (ie the frontend has cached the data)

### DIFF
--- a/app/blueprints/assessments/models/fund_summary.py
+++ b/app/blueprints/assessments/models/fund_summary.py
@@ -11,6 +11,8 @@ from app.blueprints.services.data_services import get_application_stats
 from app.blueprints.services.data_services import get_assessments_stats
 from app.blueprints.services.data_services import get_rounds
 from app.blueprints.services.models.fund import Fund
+from app.blueprints.shared.helpers import get_ttl_hash
+from config import Config
 from config.display_value_mappings import ALL_VALUE
 from config.display_value_mappings import LandingFilters
 from flask import current_app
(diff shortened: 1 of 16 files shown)


tferns (dismissed):
Looks good.  But I do wonder if we should have some script/end-point for manually clearing the caches?  Or is restarting the app in cloudfoundry good enough maybe.  

I.e could be something as simple as:

```python
@app.route('/clear_cache', methods=['POST'])
def clear_cache():
    get_funds.cache_clear()
    get_round.cache_clear()
```

Anyways, just a thought.  Don't think it's a show-stopper on your PR as we can always restart the cloudfoundry app, although not sure how "good" this is to do on prod.

```bash
cf restart funding-service-design-assessment
```

Adam-W1 (approved):
LGTM

gidsg:
@RamuniN on production we have multiple instances so couldn't guarantee a user would hit a cache instance. If we're going this route rather than using something like cloudfront to cache, may be worth moving into Redis on the environments using something like: https://flask-caching.readthedocs.io/en/latest/ which would allow us to cache in a shared backend.

RamuniN:
> @RamuniN on production we have multiple instances so couldn't guarantee a user would hit a cache instance. If we're going this route rather than using something like cloudfront to cache, may be worth moving into Redis on the environments using something like: https://flask-caching.readthedocs.io/en/latest/ which would allow us to cache in a shared backend.

Thanks @gidsg for pointing this out. I have created a ticket in JIRA https://dluhcdigital.atlassian.net/browse/FS-3584 to tackle this

PR #452, opened 2023-10-24
BAU: Make filenames unique

tferns:
- Add date & time to the filename, and the report type, so filenames are not always duplicated

### DIFF
--- a/app/blueprints/assessments/routes.py
+++ b/app/blueprints/assessments/routes.py
@@ -673,9 +673,10 @@ def assessor_export(
     en_export_data = generate_assessment_info_csv(export["en_list"])
     cy_export_data = generate_assessment_info_csv(export["cy_list"])
 
+    formatted_datetime = datetime.utcnow().strftime("%Y-%m-%d_%H-%M-%S")
     files_to_download = [
-        ("en_export_data.csv", en_export_data),
-        ("cy_export_data.csv", cy_export_data),
+        (f"{report_type.lower()}_{formatted_datetime}_en.csv", en_export_data),
+        (f"{report_type.lower()}_{formatted_datetime}_cy.csv", cy_export_data),
     ]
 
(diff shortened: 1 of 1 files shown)


app/blueprints/assessments/routes.py
```diff
@@ -673,9 +673,10 @@ def assessor_export(
     en_export_data = generate_assessment_info_csv(export["en_list"])
     cy_export_data = generate_assessment_info_csv(export["cy_list"])
 
+    formatted_datetime = datetime.utcnow().strftime("%Y-%m-%d_%H-%M-%S")
     files_to_download = [
-        ("en_export_data.csv", en_export_data),
-        ("cy_export_data.csv", cy_export_data),
+        (f"{report_type.lower()}_{formatted_datetime}_en.csv", en_export_data),
```
RamuniN, line 678: can we have `export_data` in filename? so we know what is this file about after download

tferns: `export_data` is a hardcoded string in the removed lines, not any variable - unless you meant something else?

RamuniN: I meant with the changes you made, the downloaded filename will be like `csv_2021101545_en.csv`, whereas previously it was like `en_export_data.csv`. can we make it like `export_data_2021101545_en.csv` ? Let me know if I'm assuming wrong here

tferns: Ohhh, `report_type` should be `ASSESSOR_EXPORT` or `OUTPUT_TRACKER` - so that already gives information, the filetype is always `csv` 👍

RamuniN: Ahh ok , I assumed `report_type` to be `csv` here. My bad🤐

RamuniN (approved):
LGTM

rayhan15243 (approved):

PR #458, opened 2023-11-03
add combine PR workflow

Adam-W1:
Workflow to combine dependabot PR's

### DIFF
--- a/.github/workflows/combine-prs.yml
+++ b/.github/workflows/combine-prs.yml
@@ -1,0 +1,152 @@
+name: 'Combine PRs'
+
+# Controls when the action will run - in this case triggered manually
+on:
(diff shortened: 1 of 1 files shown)


.github/workflows/combine-prs.yml
```diff
@@ -0,0 +1,152 @@
+name: 'Combine PRs'
```
aaronwilliamsv1, line 1: I'd maybe call this "Combine dependabot PRs" to be properly specific. I know you can override the branchPrefix, as it is now, but it might be an idea to remove that input and keep that as a static value. However, dependabot might change the branch prefix. Six and two threes

aaronwilliamsv1 (approved):
LGTM

PR #554, opened 2024-01-25
FS-3769 show funds in dashboard only when length>0

RamuniN:
https://dluhcdigital.atlassian.net/browse/FS-3769

### Description
- fix `show all sections` dropdown when fund length=0


tferns (approved):
Nice one

PR #558, opened 2024-02-02
FS-2713  single config file for all code formatters

RamuniN:
https://dluhcdigital.atlassian.net/browse/FS-2713

### Description
Use a single config file (`pyproject.toml`) for all the code formatters (flake8, black..)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,8 +10,5 @@ line-length = 120
 experimental-string-processing = 1
 
 [tool.flake8]
-per-file-ignores = [
-    './app/assess/routes.py:F405,F403,E731',
-]
 max-line-length = 120
 count = true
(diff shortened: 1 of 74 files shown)


pyproject.toml
```diff
@@ -6,5 +6,12 @@ authors = ["Version One"]
 license = "MIT License"
 
 [tool.black]
-line-length = 79
+line-length = 120
 experimental-string-processing = 1
+
+[tool.flake8]
+per-file-ignores = [
+    './app/assess/routes.py:F405,F403,E731',
```
srh-sloan, line 14: Could we add a comment about what is ignored here and why please?

RamuniN: Actually, this setting is moved from `.pre-commit-config.yaml`. But now this file no more exists so removing the settings

PR #571, opened 2024-02-22
fs-4158-dashboard-appearance

srh-sloan:
Updated dashboard to match design.

Note the 'submission date' column won't be populated yet as the backend changes to provide this info are yet to be done.


rayhan15243 (approved):
Looks good to me.

There is already a bug ticket to add the submission date column:
https://dluhcdigital.atlassian.net/browse/FS-4104

PR #572, opened 2024-02-22
fs-4518 adding date_submitted to fields

srh-sloan:
Using new field from assessment store to display formatted submission date


rayhan15243 (approved):
Looks good to me.

This PR also resolves bug: https://dluhcdigital.atlassian.net/browse/FS-4104

PR #582, opened 2024-03-05
Bau: fix cof eoi open date card display

RamuniN:
### Description

Fixed display of incorrect COF-EOI metric card when the application is in live mode

We have `round.opens` == `round.assessment_start` set for `COF-EOI`, the expected display metric is like below

image
but instead below one is displayed
image


adamdavies1 (approved):
LGTM

PR #586, opened 2024-03-08
modify function for `s3bucketPath` display type

RamuniN:
https://dluhcdigital.atlassian.net/browse/FS-4269

### Description
Fixed bug - file upload not showing for  `Upload asset valuation or lease agreement` question


srh-sloan (commented):
Do we have unit tests for this?

RamuniN:
> Do we have unit tests for this?
Existing unit tests already check `s3bucketPath` questions

srh-sloan:
> > Do we have unit tests for this?
> > Yes, existing unit tests already check `s3bucketPath` questions

OK but you've changed the code to do something different for a different type of answer, so you need to change the tests to cover that new scenario

PR #591, opened 2024-03-14
Fs 4268 fund summary logic

srh-sloan:
### Change description
Reviewed the logic of the Assessment Tool Dashboard (Landing) page to simplify and make it easier to debug, and easier to test.
- Removed the `FORCE_OPEN_ALL_LIVE_ASSESSMENT_ROUNDS` env var as we can mimic this with dates in the DB
- Extracted some of the logic in the fund_summary template into macros for easier testing
- Renamed fund_summary.py to round_summary.py as it was referred to by both names in different places, and is actually used per round not per fund
- Resolves FS-4268 (link display for EOI)
- Extracted the logic around what status a round is currently in to a single dedicated method and class (`RoundStatus`) to hold the results. Refactored code to use this.
- Wrote unit tests for the above

- [x] Unit tests and other appropriate tests added or updated

### How to test
Launch the assessment tool and check the round summaries display correctly for rounds in different statuses

Only UI change is the fixing of the link text for EOI (FS-4268)
Screenshot 2024-03-14 at 13 30 13


app/blueprints/assessments/models/fund_summary.py
RamuniN, line 1: Here you are deleting the file instead of renaming it. We lose file history if deleted instead of renaming. Is this what you want here?

app/blueprints/assessments/models/round_status.py
```diff
@@ -0,0 +1,48 @@
...
+from fsd_utils.simple_utils.date_utils import (
+    current_datetime_before_given_iso_string,
+)
+
+
+@dataclass
+class RoundStatus:
+    is_application_not_yet_open: bool
+    is_application_open: bool
+    is_application_closed: bool
```
RamuniN, line 19: Do both of these variables tell the same thing?  can we keep one and use `not` to check negative scenario?

app/blueprints/assessments/templates/macros/fund_dashboard_summary.html
```diff
@@ -48,90 +76,76 @@ <h3 class="govuk-body govuk_summary-card__title govuk-!-font-size-24">{{summary.
...
+                    [{'html': '<strong>QA Complete</strong>'},
+                    {'text': summary.assessment_stats.qa_complete}],
+                    [{'html': '<strong>Stopped assessments</strong>'},
+                    {'text': summary.assessment_stats.stopped}],
+                    ],
+                })
+            }}
 
         {% endif %}
+        {% if not summary.status.is_application_not_yet_open %}
```
RamuniN, line 140: Can you double-check the logic here? We are not supposed to show 'assessment tracker export' link for live rounds I think
image

RamuniN (commented):
Looks a lot better now. A few minor comments

app/blueprints/assessments/models/fund_summary.py
srh-sloan: Not intentionally! Is there a way to tell git to rename instead of delete?

app/blueprints/assessments/models/round_status.py
srh-sloan: They kind of do, but not quite. Also they can both be `False` when `is_application_not_yet_open` is true, and having them as separate properties I think makes the logic easier to read in places rather than having additional `not` clauses.

app/blueprints/assessments/models/fund_summary.py
RamuniN: yes. `git mv <original_filepath> <destination_filepath>`

app/blueprints/assessments/templates/macros/fund_dashboard_summary.html
RamuniN: I think we also hide  `export applicant information` link on Live round as there are no applications in the assess store it might crash.

app/blueprints/assessments/models/round_status.py
srh-sloan: Actually I've renamed it now to `has_application_closed` so it's only `true` after the `deadline` has passed - which is useful for the link display thing you pointed out further down

app/blueprints/assessments/models/fund_summary.py
srh-sloan: Despite that other branch to delete the file and then rebasing, it still wants to mark it as deleted

RamuniN (dismissed):

app/blueprints/assessments/models/fund_summary.py
srh-sloan: Agreed not to spend any more time on this

RamuniN (approved):

PR #595, opened 2024-03-20
fs-4294 using shared workflow to determine jobs

srh-sloan:
Updated copilot_deploy.yml to use the new shared workflow to determine which jobs to run


Tiny49 (commented):
LGTM

Tiny49 (approved):
LGTM

PR #607, opened 2024-04-03
FS-4250: Translate checkbox radio fields

tferns:
### Change description

- At the current time, the assessment service is English.  So we translate and map any Welsh radio/checkbox fields to English, for the assessors.
- This solution is admittedly quite hacky, and for future rounds we'll need to keep it up to date, thinking of a bigger fix we'd need to fix the root cause of the problem.
- Script has been written to automatically generate a mapping of Welsh to English translations for all rounds with both English and Welsh, works dynamically.

### How to test

- I tested this manually by going onto my local app and mutating the data in the database, and creating a few Welsh applications as well, the logic is very trivial and simple.  Unit tests updated to work with the new format.


srh-sloan (approved):
Very cool! My only thought is that we should store the script and the generated values in separate files - that way the history is clearer about when the script was changed and when it was just run to generate new values. So you could change the script to output to a file rather than the console?

srh-sloan (approved):
🚀

PR #656, opened 2024-07-10
Fs 3981 reporting to you tab

Adam-W1:
https://dluhcdigital.atlassian.net/browse/FS-3981

image


app/blueprints/assessments/routes.py
```diff
@@ -265,6 +266,11 @@ def fund_dashboard(fund_short_name: str, round_short_name: str):
         fund_short_name,
     )
 
+    future_applications_reporting_to_user = thread_executor.submit(
```
ksp37-dluhc, line 269: So after my change, `get_application_overviews()` should now return all the assigner / assignee data we need. All that's left to do is cross reference it with user data to get names, emails etc (which is done in `set_assigned_info_in_overview()` ). I think you could probably extend that method, or create a new one.

Adam-W1: right you are, that simplifies things nicely! ill make the changes

ksp37-dluhc (approved):

PR #670, opened 2024-08-12
FS-3987 assign user assessment

ksp37-dluhc:
### Change description
This PR covers several views to enable lead assessors to assign assessments to other lead or general assessors. Things to note:
- The actual assignments are only created after submission on the final page (assignment_overview). The views prior to this only store user selections via hidden input fields (persistent storage is not used until the final submission). 
- This development only covers the ability to assign users to assessments (un-assignment will be covered in another ticket).
- The search bar and 'Team' column are no longer needed. The optional comment field will be covered in another ticket.

### How to test
Follow the figma flow in the ticket to test the workflow and views. 

Screenshot 2024-08-13 at 11 04 14
Screenshot 2024-08-13 at 11 04 23
Screenshot 2024-08-13 at 11 45 56
Screenshot 2024-08-13 at 11 46 12
Screenshot 2024-08-13 at 11 46 39

### DIFF
--- a/app/blueprints/assessments/routes.py
+++ b/app/blueprints/assessments/routes.py
@@ -101,6 +101,8 @@ from app.blueprints.services.data_services import get_users_for_fund
 from app.blueprints.services.data_services import match_comment_to_theme
 from app.blueprints.services.data_services import submit_comment
 from app.blueprints.services.models.comment import CommentType
+from app.blueprints.services.models.fund import Fund
+from app.blueprints.services.models.round import Round
 from app.blueprints.services.models.theme import Theme
 from app.blueprints.services.shared_data_helpers import get_state_for_tasklist_banner
 from app.blueprints.shared.filters import utc_to_bst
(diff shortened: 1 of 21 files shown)


app/blueprints/assessments/routes.py
```diff
@@ -411,36 +362,493 @@ def get_sorted_application_overviews(application_overviews, column, reverse=Fals
...
+            307,
+        )
+
+    fund = get_fund(
+        fund_short_name,
+        use_short_name=True,
+        ttl_hash=get_ttl_hash(Config.LRU_CACHE_TIME),
+    )
+    if not fund:
+        return redirect("/assess/assessor_tool_dashboard/")
```
nuwan-samarasinghe, line 573: it seems this block can be generalised 



```
if form.validate_on_submit():
        return redirect(
            url_for(
                "assessment_bp.assignment_overview",
                fund_short_name=fund_short_name,
                round_short_name=round_short_name,
            ),
            307,
        )

    fund = get_fund(
        fund_short_name,
        use_short_name=True,
        ttl_hash=get_ttl_hash(Config.LRU_CACHE_TIME),
    )
    if not fund:
        return redirect("/assess/assessor_tool_dashboard/")

round = get_round(
        fund_short_name,
        round_short_name,
        use_short_name=True,
        ttl_hash=get_ttl_hash(Config.LRU_CACHE_TIME),
    )

```

app/blueprints/assessments/routes.py
```diff
@@ -411,36 +362,493 @@ def get_sorted_application_overviews(application_overviews, column, reverse=Fals
...
+        "assessment_deadline": round.assessment_deadline,
+        "round_title": round.title,
+        "fund_name": fund.name,
+        "fund_short_name": fund_short_name,
+        "round_short_name": round_short_name,
+        "is_expression_of_interest": round.is_expression_of_interest,
+    }
+
+    thread_executor = ContextAwareExecutor(
+        max_workers=4,
```
nuwan-samarasinghe, line 592: shall  we add this into the config ? and for every request are we going to create this ContextAwareExecutor ? better to have one executor for entire app other wise more ContextAwareExecutor will be created

app/blueprints/assessments/routes.py
```diff
@@ -411,36 +362,493 @@ def get_sorted_application_overviews(application_overviews, column, reverse=Fals
...
+    )
+
+    all_applications_metadata = future_all_applications_metadata.result()
+    users_for_fund = future_users_for_fund.result()
+
+    if assessor_role.lower() == "lead_assessor":
+        selectable_users = [
+            user
+            for user in users_for_fund
+            if any("LEAD_ASSESSOR" in role for role in user["roles"])
```
nuwan-samarasinghe, line 620: shall we add constant values for these ?

app/blueprints/assessments/routes.py
```diff
@@ -411,36 +362,493 @@ def get_sorted_application_overviews(application_overviews, column, reverse=Fals
...
+                "assessment_bp.assign_assessments",
+                fund_short_name=fund_short_name,
+                round_short_name=round_short_name,
+            ),
+            307,
+        )
+
+    selected_assessments = request.form.getlist("selected_assessments")
+    assessor_role = request.form.getlist("assessor_role")[0]
+    selected_user_ids = request.form.getlist("selected_users")
```
nuwan-samarasinghe, line 684: if the list is empty we might need to handle 
```request.form.getlist("assessor_role")```

File: app/blueprints/assessments/routes.py [hunk 5]
```diff
@@ -411,36 +362,493 @@ def get_sorted_application_overviews(application_overviews, column, reverse=Fals
...
+
+    form = AssignmentOverviewForm()
+
+    thread_executor = ContextAwareExecutor(
+        max_workers=4,
```
nuwan-samarasinghe, line 689: check the comment related to `ContextAwareExecutor`

app/blueprints/assessments/routes.py
ksp37-dluhc: So we don't use the ContextAwareExecutor for every request, only those that involve a lot of calls to the stores. The problem with having a single executor for the app is that all requests can/will use it, and that will introduce coupling between requests ... so if we have one bad request that blocks the executor, it may end up blocking other unrelated requests. It would also be a nightmare to debug. 

I think creating an executor per request is probably the safest approach (and it will be destroyed at the end of the request anyway).

(To be honest, I don't really like using the ContextAwareExecutor approach ... but we have to because of the microservice architecture )

app/blueprints/assessments/routes.py
ksp37-dluhc: As above.

app/blueprints/assessments/routes.py
ksp37-dluhc: As above.

app/blueprints/assessments/routes.py
ksp37-dluhc: Sure will do that.

app/blueprints/assessments/routes.py
nuwan-samarasinghe: shall we have a call about this I have some previous knowledge about java it caused an server overload but since we have ecs it can handle it but it creates unnecessary overload

ksp37-dluhc: Summary of call : `ContextAwareExecutor` doesn't automatically clean up the threads when it goes out of scope, so it is necessary to call `ContextAwareExecutor.executor.shutdown()` to ensure the necessary clean up takes place.

nuwan-samarasinghe (approved):
LGM

PR #674, opened 2024-08-20
Update get_application_assignments to default to empty list

ksp37-dluhc:
### Change description
Quick fix for get_application_assignments to return empty list instead of None. 
Also extend assign_user_to_assessment to be able to un-assign assessments.


NarenderRajuB (approved):
LGTM

PR #690, opened 2024-10-08
FS-4001: Add Assigned to you section to assessment overview

suegarner:
_Add ticket reference to Pull Request title: e.g. 'FS-123: Add content', if there is no ticket prefix with BAU_

### Change description
_A brief description of the pull request_

### How to test
_If manual testing is needed, give suggested testing steps_

### DIFF
--- a/app/blueprints/assessments/routes.py
+++ b/app/blueprints/assessments/routes.py
@@ -514,6 +514,8 @@ def assessor_type(fund_short_name: str, round_short_name: str):
     if not (selected_assessments := request.form.getlist("selected_assessments")):
         abort(500, "Required selected_assessments field to be populated")
 
+    from_form=request.form.get("from_form")
+
     selected_assessor_role = (
         request.form.getlist("assessor_role")[0]
         if request.form.getlist("assessor_role")
@@ -567,6 +569,7 @@ def assessor_type(fund_short_name: str, round_short_name: str):
         round_details=round_details,
         stats=unfiltered_stats,
         form=form,
+        from_form=from_form,
     )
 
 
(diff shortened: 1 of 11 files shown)


app/blueprints/assessments/routes.py
```diff
@@ -514,6 +514,8 @@ def assessor_type(fund_short_name: str, round_short_name: str):
     if not (selected_assessments := request.form.getlist("selected_assessments")):
         abort(500, "Required selected_assessments field to be populated")
 
+    from_form = request.form.get("from_form")
```
sfount, line 517: I wonder if something more explicit like `redirect_to_application_detail` would make the control flow read easier. I think originally `is_application` or `is_application_dashboard` which also sound like good options. 

Another option might be `action=application_overview` or an enum which would be more flexible but I suspect these are the only two that will be needed so thats likely overdoing it!

app/blueprints/assessments/routes.py
```diff
@@ -1629,6 +1646,54 @@ def application(application_id):
             )
         )
 
+    thread_executor = ContextAwareExecutor(
```
sfount, line 1649: I'm not sure the savings from the thread executor mechanism are worth the complexity or if this logic shouldn't be encapsulated lower down. I know this logic was added before this PR so all good for it to go through.

I wonder if using the `async` functionality added to flask recently (relatively!) would be a less hands on way of managing waiting for multiple calls.

sfount (commented):
Code and template are looking good! As discussed last week we thought we could cover this by adding an assertion for the default non-assigned behaviour on the existing application detail page tests first and then extend the mocks to add the scenarios where a non-lead and then lead assessor are assigned.

app/blueprints/assessments/routes.py
suegarner: Renamed to redirect_to_application_detail

sfount:
Ticket is currently blocked on design changes & access to environments - closing for combining 
 the frontends and we can support in gettin the branch pushed and PR raised against the new repo

PR #694, opened 2024-10-21
FS-4647 Visually differentiate expresssion of interest funds

sfount:
### Change description
This adds a suffix to the name of funds with the `funding_type` property "EOI".

Note that the fund name is used in many places in the assessment tool so for now this just sets the suffix in the data model, ideally this should be a presentation layer (template) concern and be done by jinja based on the funds properties. This however does the job.

Committing for opinions and consensus.


sfount:
Note that I'll spin off a separate small ticket relating to this, it will update the specific places fund names are used to respond to the funding type to make the behaviour more explicit and clear from the presentation layer.

This should work to review the desired behaviour in the meantime

srh-sloan (approved):
LGTM

PR #698, opened 2024-11-07
Read maintenance mode variables from param store

samuelhwilliams:
This will allow us to turn maintenance mode on/off with a config change and re-deploy, rather than with a code change in a PR. It also means that we can put individual environments into maintenance mode, rather than requiring all envs go into maintenance mode at once.

Would like to pair with whoever reviews this on adding the vars to each environment. Should simply be running these commands for each env:

```
copilot secret init --app fsd-assessment --name MAINTENANCE_MODE --values dev=False
copilot secret init --app fsd-assessment --name MAINTENANCE_END_TIME --values 'dev='
```

https://mhclgdigital.atlassian.net/browse/FSPT-69


srh-sloan (approved):

samuelhwilliams:
Tested in dev and works as expected ✅

PR #723, opened 2024-12-06
FSPT-119 Remove SCSS preprocess step

sfount:
We aren't importing and using the colour variables that would track
design system changes anyway. Remove the variables for now - if we start
building on the design system scss and bundling appropriately we can add
that back in.

Now that the files are just css remove the pyscss filter step. This lets
us remove the dependency.

While I'm here it doesn't look like pyopenssl is used so remove that.

---

It doesn't look like the `assets.py` build step is used anymore (in
favour of the static assets build step?) so remove it.


MarcUsher (approved):
Love a PR with sensible commits that make it easy to review 🙌

